PR #2, opened 2020-02-15
Add option to run a script when client connects or disconnects (Linux…

tekdoc:
… only for now)


aler9:
Hello, thanks for the PR, i merged it after doing two changes:
* the code was formatted
* i removed the OS check, since exec.Command() is supported by both Linux and Windows, so nothing prevents us from using this feature in Windows too, it's enough to write the script full path (i.e. `C:\folder\script.exe`)

PR #3, opened 2020-02-22
Updating readme to add rtsp-relay links

flaviostutz:


aler9:
Hello, thanks for your interest in this project and thanks for the docker image.
I don't understand why you're using both rtsp-simple-proxy and rtsp-simple-server to serve RTSP, i think it should be enough to use ffmpeg, ie:
```
ffmpeg -re -i rtsp://input -f rtsp rtsp://simple-proxy
```

Maybe you did it to solve the port problem, since a docker container doesn't have access to all the host ports, and therefore ffmpeg can't receive UDP streams. Unfortunately neither rtsp-simple-proxy can, as it uses two random ports ( https://github.com/aler9/rtsp-simple-proxy/blob/master/stream.go#L313 ). rtsp-simple-server, on the other hand, use two fixed ports.
Maybe you could hack rtsp-simple-proxy to use deterministic ports and solve this problem.

flaviostutz:
Thanks for the detailed reply!
I do have a flag to use ffmpeg instead of simple-proxy for rtsp2rtsp bridging. I think in the future I won’t use proxy indeeded, but I am doing some performance tests because low CPU usage is important in our application, and thought proxy would more efficient than restreaming with ffmpeg.
I have to expand tests because with low number of clients it was not conclusive.
What do you think?

aler9:
Hello, you're dealing with two different problems:
* containerizing an RTSP client - this is tricky, since an RTSP client listens on two random ports for receiving UDP streams, UDP is the default RTSP medium and a Docker container only exposes some ports, not all. If you want to proceed without using `--network=host`, you have to hack rtsp-simple-proxy OR ffmpeg in order to listen on fixed ports.

* Proxying streams without re-encoding and with the lowest possible CPU usage. Of course it's possible and i did it in many projects. If the source is an RTSP stream, you can simply copy the bytes from one end to the other without even touching them, and that's what rtsp-simple-proxy does. ffmpeg can do it too, with the `-c copy` flag, and is called remuxing, as opposite to re-encoding:
```
ffmpeg -re -i rtsp://source -c copy -f rtsp rtsp://server
```

The approaches are more or less equivalent, even in terms of performance. If you want to increase the performance further, you have to write something from scratch based your specific scenario.

Hope this helps.

flaviostutz:
I was thinking about using sendfile (DMA) to even avoid copying the bytes from one socket to another through CPU. Have you done this already? Could I avoid using the write/read cycle to disk and send the data directly from one socket to another through DMA?

flaviostutz:
O found something here: https://github.com/golang/go/issues/10948

flaviostutz:
Forgot to mention... thanks for the UDP random port problem clarification. I was thinking it was on 554, but makes all sense as each client needs a different port because UDP is stateless 😬. For my case I think TCP will do the job for now because it will pass through a local network communication.

PR #31, opened 2020-07-13
Fix markdown issues

x-jokay:
Fixes markdown issues found by [markdownlint](https://github.com/markdownlint/markdownlint).


aler9:
Hi, sorry but i prefer using non-standard markdown, the language is already very strict (if HTML is not used, which i don't) and i want to maintain a FAQ section with question titles in a smaller font size.

Nonetheless i'll add the sh language to the code.

x-jokay:
*... you can take a horse to water but you can't make it drink ...*

Issue #47, opened 2020-07-21
Question: Can I use runOnPublish option with dynamic url


LuiGee3471:
I'm trying to save input rtsp as another video file, but it seems I can only specify one url when I'm using ffmpeg on runOnPublish option.

Is there an way to use its own rtsp url from input as variable?

aler9:
Hello, variables inside runOnPublish / runOnRead / runOnDemand will be supported soon.

LuiGee3471:
Good to hear that. Looking forward to it.

aler9:
Hello, this should be implemented in v0.9.9 - use a configuration like this:
```yaml
paths:
  all:
    runOnPublish: ffmpeg -i rtsp://localhost:8554/$RTSP_SERVER_PATH -c copy -f mpegts myfile.ts
```

PR #52, opened 2020-08-02
First attempt at implementing wildcard paths

xieliwei:
Apologies for the duplicate earlier on, should have been less jumpy about opening an issue!

First attempt at getting this working, based on my limited understanding of RTSP and the code structure.

This patch introduces a new configuration item: `isWildcard`, that causes the configured path to be matched against any path that begins with it. The `runOnInit`/`runOnDemand`/`runOnPublish`/`runOnRead` callers have also been modified to set an environment variable RTSP_SERVER_PATH with the calling path. (For runOnInit, it will be called with the configured path, of course)

`findConfForPath()` has been modified to return an additional string - the configured path that was matched. It isn't necessary at the moment (only used to print out logs), but might be useful to have.

There are some doubts I have:
- Possibly incorrect behaviour if `isWildcard` is set for the 'all' path
- Unsure if there will be odd interactions with sourceOnDemand, thus currently there is a check to make sure that sourceOnDemand is off when isWildcard is on
- It seems like it should be sufficient to handle adding the dynamic path during the describe event, but I may be wrong
- There is a check to only destroy the dynamic path when both the producer and consumers are disconnected, but it possibly does not handle multiple clients connecting to the same path correctly
- I have only tested this for my use case, which is with dynamically generated recorder streams, and only one client per path
- The environment variable name should probably be a configuration item, and it should be possible to turn the feature off

Do let me know how I can improve the patch for acceptance, thanks!

(Addresses #47 )

### DIFF
--- a/client.go
+++ b/client.go
@@ -316,7 +316,7 @@ func (c *client) handleRequest(req *gortsplib.Request) bool {
 			return false
 		}
 
-		confp := c.p.findConfForPath(path)
+		confp, _ := c.p.findConfForPath(path)
 		if confp == nil {
 			c.writeResError(req, gortsplib.StatusBadRequest,
 				fmt.Errorf("unable to find a valid configuration for path '%s'", path))
@@ -362,7 +362,7 @@ func (c *client) handleRequest(req *gortsplib.Request) bool {
 			return false
 		}
 
-		confp := c.p.findConfForPath(path)
+		confp, _ := c.p.findConfForPath(path)
 		if confp == nil {
 			c.writeResError(req, gortsplib.StatusBadRequest,
 				fmt.Errorf("unable to find a valid configuration for path '%s'", path))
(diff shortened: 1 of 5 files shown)


aler9:
Hello, first of all thanks for your contribution.
If possible, i'd split the work in two separate PRs, since you're addressing two separate features:
* one for wildcards in paths
* one for variables in commands. Since i've already received some user feedback regarding this feature, i think it would be more versatile to provide values via string replacements instead of environment variables, since a portion of users don't know what environment variables are or don't want to use them  - in this way it would be possible to avoid writing scripts in simple cases like this:
```
runOnDemand: ffmpeg -i ... -f rtsp rtsp://localhost:8554/$path
```

where $path is a variable provided with
```
strings.ReplaceAll(cmd, "$path", path)
```

xieliwei:
Done, I've opened two new pull requests. If they're okay, do proceed to close this one.

(Unrelated to this, the latest changes to UDP handling seems to have broken something on our streaming setup fed with gstreamer rtspclientsink. I'll try to find time next week to investigate this.)

PR #54, opened 2020-08-04
implement simple handling of wildcard paths

xieliwei:
This patch introduces a new configuration item: `isWildcard`, that causes the configured path to be matched against any path that begins with it.

`findConfForPath()` has been modified to return an additional string - the configured path that was matched. It isn't necessary at the moment (only used to print out logs), but might be useful to have.

There are some doubts I have:
- Possibly incorrect behaviour if `isWildcard` is set for the 'all' path
- Unsure if there will be odd interactions with `sourceOnDemand`, thus currently there is a check to make sure that `sourceOnDemand` is off when `isWildcard` is on
- Possible weird interactions with the newly implemented feature for handling control paths (d0c2c3a58653ea7fe650c5eb851c228615149beb)
- It seems like it should be sufficient to handle adding the dynamic path during the describe event, but I may be wrong
- There is a check to only destroy the dynamic path when both the producer and consumers are disconnected, but it possibly does not handle multiple clients connecting to the same path correctly
- I have only tested this for my use case, which is with dynamically generated recorder streams, and only one client per path

Do let me know how I can improve the patch for acceptance, thanks!

### DIFF
--- a/client.go
+++ b/client.go
@@ -309,7 +309,7 @@ func (c *client) handleRequest(req *gortsplib.Request) bool {
 			return false
 		}
 
-		confp := c.p.findConfForPath(path)
+		confp, _ := c.p.findConfForPath(path)
 		if confp == nil {
 			c.writeResError(req, gortsplib.StatusBadRequest,
 				fmt.Errorf("unable to find a valid configuration for path '%s'", path))
@@ -360,7 +360,7 @@ func (c *client) handleRequest(req *gortsplib.Request) bool {
 			return false
 		}
 
-		confp := c.p.findConfForPath(path)
+		confp, _ := c.p.findConfForPath(path)
 		if confp == nil {
 			c.writeResError(req, gortsplib.StatusBadRequest,
 				fmt.Errorf("unable to find a valid configuration for path '%s'", path))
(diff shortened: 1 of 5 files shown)


aler9:
Hello, this is a good starting point, i think improvements are possible in two directions:
* you wrote "if the source is an RTSP url, sourceOnDemand must be set to 'no'" - i think RTSP sources must be disabled on wildcard paths; a wildcard path in the configuration doesn't correspond to a single path, but to an indefinite number of paths, and at the moment a source must be attached to a single path.
Of course it would be possible to change the design in a way in which RTSP sources could be attached to wildcard paths, but i don't think this feature would be useful to most users (maybe on-demand sources with dynamic urls are of some utility, but we're digging into a subject that is very far from this PR).

* Why don't directly support regular expressions in path names instead of using a wildcard? In this way we can support both
```
prefix*
```
and
```
more-complicated-prefix[0-9]{2}
```
the resulting implementation is simpler to write than the actual one.

aler9:
Implemented in a different way with 6679b02, will be available in the next release.
The difference with this PR is that there isn't a wildcard parameter, but paths that starts with a tilde are interpeted like regular expressions.
For instance, this configuration
```
paths:
  '~^(test1|test2)$':
```
will match both `test1` and `test2`, while this
```
paths:
  '~^prefix':
```
will match every path that starts with prefix.
maybe it's not as straightforward as the original proposal, but is much more versatile.

aler9:
alternative implementation added in v0.10.0

PR #55, opened 2020-08-04
expose request path to runOn commands as an environment variable

xieliwei:
This patch modifies the `runOnInit`/`runOnDemand`/`runOnPublish`/`runOnRead` callers to set an environment variable with the calling path. (For `runOnInit`, it will be called with the configured path, of course)

A `pathEnvVariable` configuration item is also introduced to specify the name of the environment variable to be used. Its default value is `RTSP_SERVER_PATH` and there is sanity checking to ensure that the provided name is valid (alphanumeric + underscore, no leading digit).

(Addresses #47)


aler9:
merged, thanks, but i'll remove the PathEnvVariable option since 99.99% of users won't need it.

Issue #79, opened 2020-09-07
Implement external commands for Windows


Fruchtzwerg94:
## Which version are you using?

v0.9.13

## Describe the problem

Adding a custom command like
`runOnPublish: ffmpeg -i rtsp://localhost:8554/$RTSP_SERVER_PATH -b:a 64k -f rtsp rtsp://localhost:8554/my/stream`
to the configuration file does fail on windows:

> ERR: exec: "/bin/sh": file does not exist

I think the reason is, /bin/sh should not be used on windows devices. The following two lines should be adapted to support different plattforms:

https://github.com/aler9/rtsp-simple-server/blob/79d0f2d053b4f5f6d1f1309652e49f061d580c95/path.go#L66
https://github.com/aler9/rtsp-simple-server/blob/1065fa010ba6832ef1a6674a0336de159eb139c5/client.go#L232

aler9:
Hello, you're perfectly right, command should works on all platforms, i just did not have the time to do it.
Since the feature is simple to write, PRs are welcome.

Fruchtzwerg94:
Done. See  #80

hyperfraise:
Hi. This happens on my Ubuntu machine. Whatever the command I put in runOnDemand, I have the same error.

Fruchtzwerg94:
Done.

PR #80, opened 2020-09-07
Implement all external commands for Windows

Fruchtzwerg94:
See #79
Implement all external commands for Windows


aler9:
Hello, thanks, this has been merged, but i'll make RTSP_SERVER_PATH also available through an environment variable.

Fruchtzwerg94:
Sure, looks good. Thanks for your reply.

PR #97, opened 2020-10-05
Kill processes instead of interrupting them

Fruchtzwerg94:
Kill custom commands instead of just interrupting them.
Why?:

- Interrupting not supported on Windows
- Libraries like VLC are not interrupting gracefully always. If a connection gets closed, the interrupt takes too much time to be executed until a potential new connection gets established. This leads to two processes of the same type are running, which can cause resource problems (like audio outout, ports, ...).

Just killing the processes is not the fine way, but more stable / reliable in this case.

### DIFF
--- a/client.go
+++ b/client.go
@@ -206,7 +206,7 @@ func (c *client) run() {
 	}
 
 	if onConnectCmd != nil {
-		onConnectCmd.Process.Signal(os.Interrupt)
+		onConnectCmd.Process.Kill()
 		onConnectCmd.Wait()
 	}
 
@@ -914,7 +914,7 @@ func (c *client) runPlay() bool {
 	}
 
 	if onReadCmd != nil {
-		onReadCmd.Process.Signal(os.Interrupt)
+		onReadCmd.Process.Kill()
 		onReadCmd.Wait()
 	}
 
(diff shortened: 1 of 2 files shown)


aler9:
Hello, thanks for reporting the issue. I finally got the time to turn on a Windows partition, which i never use, wait for the automatic updates and check the issue.

I didn't noticed that Go doesn't support sending SIGINT to Windows processes, probably SIGINT doesn't even exist in Windows, i don't remember since the last time i wrote something native for Windows was at eleven years old with Visual Studio 2003 (then it was only Cygwin and Python). Therefore, Kill() must be used on Windows.

At the same time, SIGINT is very handy in Linux, since it allows to write scripts that can handle both the start and the end of an event; for instance, this script
```
#!/bin/sh

trap 'echo stopped' SIGINT

echo started

while true; do sleep 86400; done
```
Will print `started` when it is started, and `ended` when it is closing.

If you need to kill a process on Linux too, it's possible to use a wrapper like this:
```
#!/bin/sh

otherprogram &
P=$?

trap 'kill -9 $P' SIGINT

wait
```

So:
* i'll switch to Kill() on Windows
* i'll keep the current behavior on Unix

Fruchtzwerg94:
Sounds good for me, thanks for your response. So I will close the pull-request - okay?

aler9:
feature added in v0.10.1

Issue #98, opened 2020-10-06
Support overriding properties from flags or environment variables


l-lin:
## Which version are you using?

v0.10.0

## Which operating system are you using?

Linux

## Enhancement request

It seems there was a time when the rtsp-simple-server could be configured via flags (for example in the #17 comment). However it's now only configurable from a yaml file (from this [improvement](https://github.com/aler9/rtsp-simple-server/commit/731cd9830e115cc7583b1579071e168c8063127f)).

I was wondering if you plan to (re-)support overriding the properties from another source than the configuration file (e.g. flags, environment variables, ...)? What do you think about https://github.com/spf13/viper?

---

My use case is that I want to spawn dynamically X rtsp-simple-servers in a single server.
For example, if I want to spawn 100 rtsp-simple-services, I could do something like:

```bash
#!/usr/bin/env bash

set -e

for i in {0..100}; do
  rtsp_port=$(expr $i + 10000)
  rtp_port=$(expr $i + 8000)
  rtcp_port=$(expr $i + 9000)
  rtsp-simple-servers --rtsp-port="${rtsp_port}" --rtp-port="${rtp_port}" --rtcp-port="${rtcp_port}"
done
```

But I can't imagine creating 100 rtsp-simple-server.yml files. It's not quite scalable.

Note: I can always mitigate my issue by creating 100 docker containers by mapping the default ports 8554, 8000 and 8001 to the host desired ports.

aler9:
feature added in v0.10.1

PR #101, opened 2020-10-11
feat(#98): add flag and env variable support

l-lin:
This pull request adds support to override properties from flags and environment variables.

I did not use [spf13/viper](https://github.com/spf13/viper) like I suggested in #98, because there is a limitation on this library as viper configuration keys are case __insensitive__, whereas the properties used in the `rtsp-simple-server.yml` are in camelCase. 

So in order to not have a breaking change, I used instead [knadh/koanf](https://github.com/knadh/koanf) to read configuration from different sources, [spf13/pflag](https://github.com/spf13/pflag) to support POSIX/GNU-style --flags and [AdamSLevy/flagbind](https://github.com/AdamSLevy/flagbind) to bind flags from a struct (in order to avoid writing property keys twice).

I completely removed kinpin as we could parse the command line flags & arguments with spf13/pflag library.

Some points to note:

- I could not make it override the `paths` property as it's a `map[string]*pathConf`, which does not seem to be supported by the libraries I used
- there is still a limitation on koanf as we cannot provide string slices from the environment variables: see https://github.com/knadh/koanf/issues/33
- this pull request brings 3 new external dependencies, not sure what's your policy on it (licenses, binary as small as possible, etc...)


aler9:
Hello, thanks for your contribution, unfortunately this implementation lacks support for maps and slices, making impossible to override part of the configuration, adding a problem (and a lot of issues from newcomers).
Maps and slices were the reason i was forced to switch to a configuration file instead of keeping the command line flags.
Since this needs to be implemented ASAP, due to the hot-reloading feature that i need to add after this, i'll see how Grafana does it and implement it in the same way (Grafana allows to override maps by using variables like PREFIX_FIRSTKEY_SECONDKEY).

l-lin:
Alright, I'll close this pull request then. Thanks for your feedback :+1:

aler9:
feature added in v0.10.1

PR #169, opened 2020-12-28
systemd improvements

gherlein:
make systemd restart on failure, add some make targets, update README

Noticed changes picked up by go automatically around go.sum and go.mod.  Generally those get checked in, but you may object.  If you do I can pull them out.


gherlein:
error - will resubmit

PR #170, opened 2020-12-28
added make targets and systemd restart

gherlein:
make systemd restart on failure, add some make targets, update README


aler9:
sorry but

* makefile is highly subjective and needed as it is for maintenance purposes
* same for .gitignore
* watchdog is disabled by default on most system services and is not useful to everybody

PR #181, opened 2021-01-12
improve config file watcher to support k8s configmap reloads

ey3ball:
improve config file watcher to support k8s configmap reloads

The config file watcher assumed a simple configuration where the
existing configuration file is overwritten by a text editor.

In practice this does not detect some more complex configuration change
scenarios such as :
* The configuration file is behind a symlink and the symlink is updated
* The configuration file is behind a double symlink (k8s configmap
  update)
* The configuration file is not updated but removed the re-created

In order to fix these cases :
* Watch the configuration file parent directory instead of the file
itself, this lets us grab events event if the file was removed and a new
file is used.
* In addition to read/write event matching, watch for any change in
the configuration file path. This handles the symlink case where the
file itself hasn't changed but its location did

### DIFF
--- a/internal/confwatcher/confwatcher.go
+++ b/internal/confwatcher/confwatcher.go
@@ -14,7 +14,8 @@ const (
 
 // ConfWatcher is a configuration file watcher.
 type ConfWatcher struct {
-	inner *fsnotify.Watcher
+	inner       *fsnotify.Watcher
+	watchedPath string
 
 	// out
 	signal chan struct{}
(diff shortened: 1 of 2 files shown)


aler9:
Thanks, i'll add some unit tests to /pkg/confwatcher, compare the cases covered by the current version with the cases covered with this PR and then i'll try to merge.

ey3ball:
Ok great !

If it's any help for the test cases, what I wanted to cover specifically is this case where rtsp-simple-server.yml is managed by kubernetes :
```
root@k3s-internal-3:/config# ls -alh
total 12K
drwxrwxrwx 3 root root 4.0K Jan 12 16:50 .
drwxr-xr-x 1 root root 4.0K Jan 12 19:06 ..
drwxr-xr-x 2 root root 4.0K Jan 12 16:50 ..2021_01_12_16_50_45.977586658
lrwxrwxrwx 1 root root   31 Jan 12 16:50 ..data -> ..2021_01_12_16_50_45.977586658
lrwxrwxrwx 1 root root   29 Jan 12 16:46 rtsp-simple-server.yml -> ..data/rtsp-simple-server.yml
```

What happens when kubernetes updates the config file is basically this :
* a new directory is created (eg: 2021_01_13_xx) and receives the new config file
* the ..data symlink is updated (can be simulated with ln -snf ..2021_01_13_xx ..data

For what it's worth I also tested the following scenario with a regular file, which previously went undetected as well
* $ ./rtsp-simple-server
* $ rm ./rtsp-simple-server.yml
* $ cp rtsp-simple-server-backup.yml rtsp-simple-server.yml

aler9:
all good, thank you very much :)

PR #190, opened 2021-01-21
fix onDemand command not getting stopped on unix/linux

### LINKED ISSUE #185: on demand command does not stop as expected?

v0.13.2

OS

- [x] Linux
- [ ] Windows
- [ ] macOS

Architecture

- [ ] amd64
- [x] arm64
- [ ] arm7
- [ ] arm6

## Describe the issue

It seems as if on demand commands will not be stopped as expected. I do start ffmpeg and pipe its output into gstreamer (which than provides the rtsp-stream) for testing as an onDemand path. The server starts the command just fine but when the client disconnects (and CLI output says `on demand command stopped`) both ffmpeg and gstreamer processes are still running and will only get terminated when rtsp-simple-server is stopped.

1. configure a path with `runOnDemand`
2. start the server
3. start a client requesting the path
4. stop the client
5. wait for the server to ouput `on demand command stopped`
6. check if command was really stopped

dlueth:
resolves #185

### DIFF
--- a/internal/externalcmd/cmd_unix.go
+++ b/internal/externalcmd/cmd_unix.go
@@ -33,7 +33,7 @@ func (e *Cmd) runInner() bool {
 
 	select {
 	case <-e.terminate:
-		syscall.Kill(-cmd.Process.Pid, syscall.SIGTERM)
+		syscall.Kill(-cmd.Process.Pid, syscall.SIGINT)
 		<-cmdDone
 		return false
 
--- a/internal/externalcmd/cmd_unix.go
+++ b/internal/externalcmd/cmd_unix.go
@@ -5,6 +5,7 @@ package externalcmd
 import (
 	"os"
 	"os/exec"
+	"syscall"
 )
 
 func (e *Cmd) runInner() bool {
@@ -15,6 +16,7 @@ func (e *Cmd) runInner() bool {
 		"RTSP_PORT="+e.env.Port,
 	)
 
+	cmd.SysProcAttr = &syscall.SysProcAttr{Setpgid: true}
 	cmd.Stdout = os.Stdout
 	cmd.Stderr = os.Stderr
 
@@ -31,7 +33,7 @@ func (e *Cmd) runInner() bool {
 
 	select {
 	case <-e.terminate:
-		cmd.Process.Signal(os.Interrupt)
+		syscall.Kill(-cmd.Process.Pid, syscall.SIGTERM)
 		<-cmdDone
 		return false
 


aler9:
Hello, this PR does two things:

* it replaces SIGINT with SIGTERM
* it kills subprocesses too

Regarding the first point, there are a lot of custom commands around that are built around SIGINT, that would stop working if this would get merged. for instance

```sh
#!/bin/sh -e
trap "echo publisher disconnected" INT
echo publisher connected
```

Therefore: the default signal can't be changed, but if you want you can implement a parameter in the configuration (for each custom command) to set the Kill signal to one of

* sigint (default)
* sigterm
* sigkill

Regarding the 2nd point, i've looked around in other servers, and

* nginx-rtmp-module doesn't kill subprocesses, it just sends a signal to the main process (https://github.com/arut/nginx-rtmp-module/blob/791b6136f02bc9613daf178723ac09f4df5a3bbf/ngx_rtmp_exec_module.c#L681)
* nginx does the same thing

In my opinion is better to write commands that, after receiving SIGINT/SIGTERM, close their subprocesses in an orderly way, for instance, i've read that you are piping gstreamer to ffmpeg - instead of inserting the pipe directly into the server configuration, write a script that wraps it

```
#!/bin/sh -e
trap 'kill $(jobs -pr)' INT TERM
gstreamer ... | ffmpeg
```

dlueth:
@aler9 Just double checked from my side:

```
#!/bin/bash
set -e

trap 'kill $(jobs -pr)' INT TERM
dumpstream | gstreamer
```

I did already try exactly this (any many other variants) before and what it does is that server logs `on demand command stopped` but the dumpstream & gstreamer keep (both) running in the background until I CTRL+C the server. The trap than (after CTRL+Cing the server) outputs `kill: usage: kill [-s sigspec | -n signum | -sigspec] pid | jobspec ... or kill -l [sigspec]`

dlueth:
OK, going to close this, had a misunderstanding of what should happen obviously.

Just for reference: I also managed to handel the signals in my bash script on Ubuntu. Had to use

```
$(ps -o pid= --ppid $$)
```

as `$(jobs -pr)` did not work

Classsic:
Hi, I still have problems when launch rtsp-simple-server from script, the SIGINT is not received for the child process, I think is related to this : [https://stackoverflow.com/questions/44000795/signal-not-received-by-bash-script-run-using-system3](https://stackoverflow.com/questions/44000795/signal-not-received-by-bash-script-run-using-system3)
When I launch the server from my shell, all works as expecting, but when launch from script, I can´t get the child process killed.

PR #235, opened 2021-03-13
add ability to configure metrics port

lionelnicolas:
Port could be specified by using either `metricsPort` config file entry or `RTSP_METRICSPORT` environment variable.


aler9:
ookay thanks!

aler9:
added in v0.15.1

PR #441, opened 2021-07-04
Fix CORS issues with HLS player using authorization

dorinclisu:
Solves https://github.com/aler9/rtsp-simple-server/issues/415


dorinclisu:
This is my first commit in Golang and I didn't know how to update the test case to cover this modification.
But it is tested in practice with HLS player in javascript app.

In theory we could also have the option `hlsAllowHeaders` in config.yaml which would default to `*` (current behavior of this PR)  but I for one don't need such fine granularity.

aler9:
Perfect, thank you very much

PR #442, opened 2021-07-04
added new metric

naman108:
this branch has added a new metric allowing for the user to view the current paths being published to

### DIFF
--- a/internal/metrics/metrics.go
+++ b/internal/metrics/metrics.go
@@ -5,7 +5,9 @@ import (
 	"io"
 	"net"
 	"net/http"
+	"reflect"
 	"strconv"
+	"strings"
 	"sync/atomic"
 	"time"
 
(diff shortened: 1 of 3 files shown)


naman108:
this pull is for v0.16.1 because we weren't able to support multiple publishers on higher versions and even here we have some issue. This pull was created to support the FreeTAKServer project for drone streams and mobile phone streams. any help would be greatly appreciated in terms of addressing the issues with multiple publishers.

shalomcrown:
I am also interested in this change, I need per-publisher incoming byte counts.

aler9:
Hello, thanks for the contribution, i'm waiting to merge this since i'm working on an HTTP API that provides the same feature. If there is any problem during the development, this will get merged.

aler9:
A PR to implement an HTTP API is finally available here #483 - the API provides a list of all paths, publishers and readers, therefore it fully covers the features proposed by this PR.

aler9:
an HTTP API is available in v0.17.0

naman108:
great thank you!

Issue #470, opened 2021-07-22
Invalid Path Name for RTSP URL


frjol:
Using the aler9/rtsp-simple-server docker image on alpine 3.12

For our testing we would like to simlulate a specific camera and noticed that we cannot replicate one manufacturer since it uses rtsp uri containing file extension which clashes with the validation that is cannot contains '.'. 

We send a RTSP DESCRIBE rtsp://<ip>/manufacturer-media/media.amp
We can see this error in the logs
`2021/07/22 14:35:30 I [3/0] [RTSP] [conn 172.23.0.1:48342] ERR: invalid path name: can contain only alfanumeric characters, underscore, minus or slash (manufacturer-media/media.amp)`

Would it be possible to remove the validation of this kind of url ?

According to the specification it should be allowed -> https://datatracker.ietf.org/doc/html/rfc3986#appendix-A

From a quick look at the code it seems it could be removed here
https://github.com/aler9/rtsp-simple-server/blob/66694640aca6dadb3b5afe9ddc5ad9e1670f7ac0/internal/conf/path.go#L61

I assume other things have to be taken in consideration

Thanks

aler9:
fixed by #471

PR #471, opened 2021-07-22
Allow dot and tilde as part of the RTSP URL

frjol:
Issue #470
Allow dot and tilde as part of the RTSP URL
Update error message


aler9:
merged, thanks

frjol:
Thanks for the approval. It's very much appreciated. 
Do you have any timeframe when this will be available in the official docker image?

aler9:
added in v0.17.0

PR #472, opened 2021-07-23
feat: allow use of regex named capture groups for paths

fionera:
That way you can use this as a relay without configuring too many paths in the config


aler9:
This can't be done for technical reasons: a source can be on-demand and can also be not on-demand: in the former case, this PR works, in the latter this PR doesn't work, since it would make all sources on-demand.

PR #509, opened 2021-08-17
README: Simplify GStreamer stream playback command

philn:
Nowadays `gst-play-1.0` hides away the complexity of the pipeline. We only need to pass a valid URI to it.


aler9:
Effectively this approach works with streams that contain video and audio, only video and only audio, while the actual one works only with streams that contain video and audio. Approved, thanks

PR #707, opened 2021-11-20
ERR: sprop-parameter-sets is missing (96 profile-level-id=4D0014;pack…

BenLocal:
In some rtsp stream, sdp may not have sps, pps data, it was passed in the NAULS data
When playing the rtmp stream, the following error will appear：

ERR: sprop-parameter-sets is missing (96 profile-level-id=4D0014;packetization-mode=0)
[https://github.com/aler9/gortsplib/issues/88](url)

### DIFF
--- a/internal/core/rtmp_conn.go
+++ b/internal/core/rtmp_conn.go
@@ -300,6 +300,9 @@ func (c *rtmpConn) runRead(ctx context.Context) error {
 	var videoStartPTS time.Duration
 	var videoDTSEst *h264.DTSEstimator
 	videoFirstIDRFound := false
+	var videoSPS []byte
+	var videoPPS []byte
+	sendVideoSPSAndPPS := false
 
 	for {
 		data, ok := c.ringBuffer.Pull()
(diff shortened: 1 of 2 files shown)


aler9:
Thanks for the contribute, but i think that the approach is wrong: 

* RTMP's H264DecoderConfig config must be written once, at the beginning of the stream, not when SPS and PPS become available
* Your procedure is applied for every reader, while it's better to save SPS and PPS once for every publisher
* this fix applies only to RTMP, while HLS also needs SPS and PPS

In my opinion, we should add an intermediate state to a RTSP source, before the source state is set to ready, in which we check if SPS and PPS are not present, and if they are not, we extract them from the NALUs, save them, and afterwards we set the source state as ready. In this way, SPS and PPS are available to any following reader, independently from the reader type (HLS muxer, RTSP reader, RTMP reader) and count.

aler9:
an alternative implementation has been added in c7b2ae8

aler9:
alternative implementation added in v0.17.10

PR #714, opened 2021-12-01
rtsp_session: log error if config parsing fails

tmatth:
Sample output:
```
2021/11/30 20:34:56 INF [RTSP] [conn 127.0.0.1:38634] closed (H264 track 1 is not valid: sprop-parameter-sets is missing (96 profile-level-id=42e01f;packetization-mode=1))
```


aler9:
merged, thanks

PR #764, opened 2022-01-11
External authentication support send url raw query

BenLocal:
Do some dynamic authentication, such as token


aler9:
merged, thanks!

PR #858, opened 2022-02-28
m3u8 relative url fixed (starting with // and with query string)

vgavro:
Urls starting with // and urls with query string parsed improperly, this pull request fixes this.

### DIFF
--- a/internal/hls/client.go
+++ b/internal/hls/client.go
@@ -176,7 +176,8 @@ func (c *Client) runInner() error {
 						c.tracksMutex.RLock()
 						defer c.tracksMutex.RUnlock()
 						c.onVideoData(pts, nalus)
-					})
+					},
+					c.parent)
 
 				go func() { errChan <- c.videoProc.run() }()
 			}
--- a/internal/hls/client.go
+++ b/internal/hls/client.go
@@ -39,8 +39,8 @@ type clientAllocateProcsReq struct {
 	res chan struct{}
 }
 
-// ClientParent is the parent of a Client.
-type ClientParent interface {
+// ClientLogger allows to receive log lines.
+type ClientLogger interface {
 	Log(level logger.Level, format string, args ...interface{})
 }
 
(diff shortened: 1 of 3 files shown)


vgavro:
Looks like https://pkg.go.dev/net/url#URL.ResolveReference is better way to do this. I'll test it and change implementation in near future.

vgavro:
> Looks like https://pkg.go.dev/net/url#URL.ResolveReference is better way to do this. I'll test it and change implementation in near future.

done

aler9:
Hello, thanks for this contribution, using `ResolveReference()` is an elegant way to solve all relative paths-related issues.

Before merging, can you post the content of a `.m3u8` playlist that contains relative URLs that couldn't be parsed before this patch? i need to add them to unit tests. Thanks

vgavro:
> Before merging, can you post the content of a `.m3u8` playlist that contains relative URLs that couldn't be parsed before this patch? i need to add it to unit tests. Thanks

Examples:
http://live.ren.tv/ren/stream1/playlist.m3u8
http://live.ren.cdnvideo.ru/ren/stream1/chunklist.m3u8

NOTE: cdnvideo.com (also cdnvideo.ru) is pretty popular in post-USSR countries service for online publishing, which is now blocked for ukrainian IPs because of war, before that there were some ukrainian tv-channels (now deleted by service).
P.S. (not relevant) Thanks for supporting Ukraine!

aler9:
I merged the fix and made sure the stream you provided is compatible with the server. HLS is implemented in multiple ways and there aren't many publicly-available sources, so it's really easy to encounter bugs. 

In my town the Ukranian flag is raised above the city hall, we're preparing to welcome and host refugees, and we've already sent a couple of trucks to the country:
https://www-mbnews-it.translate.goog/2022/03/manifestazioni-per-la-pace-in-centinaia-i-brianzoli-in-piazza-ecco-le-prossime-date/?_x_tr_sl=it&_x_tr_tl=en&_x_tr_hl=it&_x_tr_pto=wapp

PR #883, opened 2022-03-15
Fixed example with GStreamer usage in README

lkepka:
Added 'location' parameter to rtspsrc element in example in 'TCP transport' section


aler9:
merged, thanks!

PR #923, opened 2022-04-21
rtsp_source: set UA to rtsp-simple-server <version>

tmatth:
Otherwise the default is gortsplib, which is fine albeit less descriptive.


tmatth:
> Hello, while this surely makes debugging easier, on the other hand exposing a server version is a well known security issue, that allows attackers to easily find vulnerable servers by scanning the internet. All popular servers don't expose their version, and we don't want to facilitate the creation of botnets.

Typically web servers (e.g., nginx via the `server_tokens` option) allow this to be configurable (and by default, does expose its server version FWIW). Would it be acceptable to make this an option? Or just changing it to `rtsp-simple-server` (without a version) instead of `gortsplib` would I think be an improvement.

PR #938, opened 2022-05-05
Implement Low-Latency HLS

aler9:
TODO:

[ ] audio support
[ ] iPhone support (probably requires HTTP/2)


JIEgOKOJI:
lhls muxer generates very small part segments #EXTINF:1.999999968,
seg3.mp4
#EXT-X-PART:DURATION=0.020833333,URI="part388.mp4",INDEPENDENT=YES

aler9:
> lhls muxer generates very small part segments #EXTINF:1.999999968, seg3.mp4 #EXT-X-PART:DURATION=0.020833333,URI="part388.mp4",INDEPENDENT=YES apple not recomended segments smaller than 200ms, playback is unstable and corrupted even in hls.js

i've added the option `hlsPartDuration` that allows to tune the part duration.

JIEgOKOJI:
now it works great. Also found that if stream has no b-frames lhls will never playback in hls.js

hhackbarth:
Great to see that LL-HLS is on the way. I would be happy to participate in the test as well. Can you provide a Windows build for this branch?

aler9:
added in v0.19.0 🎆

PR #1039, opened 2022-07-14
Add example for overriding array parameter with env var

farshidtz:
I figured this out by looking at the original code added by https://github.com/aler9/rtsp-simple-server/commit/5ca500504064fe389eeec02106ec4d31a50765fb.

The code is now at [internal/conf/env.go](https://github.com/aler9/rtsp-simple-server/blob/main/internal/conf/env.go) and the slice processing part has been removed in https://github.com/aler9/rtsp-simple-server/commit/0d4d81c3336243ced35a0e48cd71aabbf0c4f560. Not sure why it was removed, but setting the value as comma-separated list still works.

### DIFF
--- a/README.md
+++ b/README.md
@@ -165,6 +165,11 @@ There are 3 ways to change the configuration:
    ```
    RTSP_RTSPADDRESS="127.0.0.1:8554" ./rtsp-simple-server
    ```
+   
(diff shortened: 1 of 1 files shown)


aler9:
merged, thanks

PR #1057, opened 2022-07-28
Add native support for the Raspberry Pi camera

### LINKED ISSUE #1010: Using libcamera-vid
## Describe the feature

Description

I'm trying to inject a raspberry pi camera (h264) stream into FFMPEG from within the `rtsp-simple-server.yaml` file. When running this from outside the container this works as expected, the stream is being forwarded to the rtsp server, and I can SETUP a connection from VLC.

    libcamera-vid -t 0 --inline -o - | ffmpeg -i pipe: -c copy -f rtsp rtsp://localhost:$RTSP_PORT/$RTSP_PATH

However when trying to run this from within the config, I'm getting a series of `exit code 0`.

    cam:
      runOnInit: libcamera-vid -t 0 --inline -o - | ffmpeg -i pipe: -c copy -f rtsp rtsp://localhost:$RTSP_PORT/$RTSP_PATH
      runOnInitRestart: yes

Any thoughts? I looked at the examples here (https://github.com/aler9/rtsp-simple-server#from-a-raspberry-pi-camera) but this is no longer applicable I assume (dont work with latest version)

### LINKED ISSUE #1011: Implement native support for the Raspberry Pi Camera
## Describe the feature

The method for using the Raspberry Pi camera with rtspss [described in the README](https://github.com/aler9/rtsp-simple-server#from-a-raspberry-pi-camera) doesn't work anymore due to a Raspberry Pi OS upgrade and the switch to 64 bits arch.

An optimal solution consists into implementing native support for the Raspberry Pi camera directly into server, by interfacing with the native API.

Steps:

1. find a way to interact with dynamic libraries. This is currently impossible since the server is compiled without clib support in order to improve compatibility. We have to find a way to interact with dynamic libraries and at the same time keep compatibility as broader as possible. Maybe the server can be released with two executables, one for generic aarch 64, and the other for the Raspberry Pi OS aarch 64.

2. interact with camera APIs / [MMAL interface](https://github.com/raspberrypi/userland/blob/master/interface/mmal/mmal.h) in order to extract H264 and make it available.

aler9:
Fixes #1011

Fixes #1010

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,3 +1,6 @@
 # do not add .git, since it is needed to extract the tag
 
+/tmp
 /release
+/coverage*.txt
+/apidocs/*.html
--- a/Makefile
+++ b/Makefile
@@ -2,6 +2,8 @@
 BASE_IMAGE = golang:1.18-alpine3.15
 LINT_IMAGE = golangci/golangci-lint:v1.45.2
 NODE_IMAGE = node:16-alpine3.15
+RPI32_IMAGE = balenalib/raspberrypi3:buster-run
+RPI64_IMAGE = balenalib/raspberrypi3-64:buster-run
 
 .PHONY: $(shell ls)
 
@@ -162,17 +164,6 @@ apidocs-lint:
 	docker run --rm -v $(PWD)/apidocs:/s -w /s temp \
 	sh -c "openapi lint openapi.yaml"
 
-define DOCKERFILE_RELEASE
-FROM $(BASE_IMAGE)
-RUN apk add --no-cache zip make git tar
-WORKDIR /s
-COPY go.mod go.sum ./
-RUN go mod download
-COPY . ./
-RUN make release-nodocker
-endef
-export DOCKERFILE_RELEASE
-
 define DOCKERFILE_APIDOCS_GEN
 FROM $(NODE_IMAGE)
 RUN yarn global add redoc-cli@0.13.7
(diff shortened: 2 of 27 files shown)


cedricve:
@aler9 any idea when this will be made available on Docker hub?

aler9:
@cedricve the Docker version doesn't support the RPI camera natively, but the standard version does. Just take a look at the README in order to try the feature.

cedricve:
ok thank you just wondering what this PR fixes then? Is it just support for libcamera (but outside a docker container)?

Kind regards,
Verstraeten Cédric

aler9:
This PR allows the serve to interact with libcamera natively, skipping all the Gstreamer stack. GStreamer is not needed anymore.

The Docker version may come in the future, but I have to check whether the camera can be accessed from a container.

cedricve:
it does, is you inject the devices or make it previleged. 

Kind regards,
Verstraeten Cédric

olokos:
Looks awesome, good job!

https://github.com/aler9/rtsp-simple-server/pull/1057/files#diff-fc943529e712158406c15ccb0d56f2afcd02a8d98a193c7623885e3d33366abdR4

I was wondering about those parameters, are there plans to support all of the native parameters with 1:1 names?
There's around 40 parameters or more for `libcamera-vid --help`

I was thinking, maybe instead of hardcoding parameters for libcamera-vid, some kind of wrapper approach could be used instead?

Right now it seems like most of true native Raspberry Pi Camera features/parameters are in fact unavailable, to some degree, but I might be completely wrong. :)


IIRC the bitrate is VBR by default for libcamera

olokos:
In terms of accessing host device from within docker container, that is possible!
https://stackoverflow.com/questions/24225647/docker-a-way-to-give-access-to-a-host-usb-or-serial-device

aler9:
I opened two enhancement request in order to track the two features requested by @cedricve and @olokos :

#1110 Allow to use the Raspberry Camera with Docker

#1111 Allow to set additional parameters of the Raspberry Pi Camera

Discussion will be moved there.

PR #1060, opened 2022-07-30
add failing LL-HLS test

Curid:
Video stops playing after 3 days.


aler9:
Hello, PRs are used to send patches in order to fix issues. Adding a test without a fix is useless, since continuous integration would fail onward. Please open an issue and describe the problem.

Issue #469, opened 2021-07-20
Add HLS support for h265 rtsp.


Prakash2101:
## Describe the feature

Description
The Server support HLS for h264 rtsp but when I tried using h265 rtsp it is unable to create .m3u8 file. I think for HLS it doesn't have h265 decoder.

please add HLS support for h265 rtsp.

robusoft-tangye:
I tested with a H.265 video file today.
Use VLC media player played url like rtsp://my-server-ip:8554/s1 steam, video and audio both OK.
But use Chrome or Firefox to access url like http://my-server-ip:8888/s1/ , video is black, audio is OK.

aler9:
Hello, the problem with HLS and H265 is that most browsers don't support H265. The only exception is Safari on iOS and macOS:

https://caniuse.com/?search=hevc

Therefore, even if this is implemented, it would benefit few users.

Prakash2101:
@aler9, 
Not an issue, can you please add this feature.

1021683053:
> 您好，HLS 和 H265 的问题是大多数浏览器不支持 H265。唯一的例外是 iOS 和 macOS 上的 Safari：
> 
> https://caniuse.com/?search=hevc
> 
> 因此，即使实现了这一点，也不会使少数用户受益。

Web can use H.265 decoder like this https://github.com/numberwolf/h265web.js
or https://github.com/ffmpegwasm/ffmpeg.wasm

aler9:
@1021683053 a WASM H265 decoder is very interesting, but it requires too much CPU to produce quality results.

Web streaming is heavily influenced by browser support, and if browsers continue not to support H265, transcoding from cameras to browsers can't be done. If you want to increase throughput or quality, you have to perform a server-side re-encoding from H265 to VP9 and use WebRTC, or from H265 to AV1 and use Media Source Extensions.

mihailj:
Hello Alessandro,

Chromium-based browsers (including Edge) have enabled by default h265 decoding for **platforms with hardware support**.

https://bugs.chromium.org/p/chromium/issues/detail?id=460703

See also here the updated table:

https://caniuse.com/?search=H265

This is quite a game-changer so it would be great if you can reconsider implementing HLS support for h265 codec.

Thanks,
Mihail

gitslav:
Hi @aler9. This project is an amazing work :)
[...]

PR #1242, opened 2022-11-11
Support reading with WebRTC

### LINKED ISSUE #566: Add support for webRTC streaming
The only things missing in current project is webRTC which will provide less than 0.5s latency for live streaming.
As you may be aware of hls or ll-hls has at least 1-2s delay for live streaming.

Regards
Gaurav Choubey

aler9:
Fixes #566

### Roadmap

* [x] read H264 tracks
* [x] read Opus tracks
* [x] read PCMA tracks
* [x] read PCMU tracks
* [x] read VP8 tracks
* [x] read VP9 tracks
* [x] authentication
* [x] API
* [x] metrics
* [x] TURN servers
* [x] find a way to generate and store a TLS certificate if the default one is missing


brunobaraobr:
Hello, first I would like to thank you for the beautiful work you have done, it is simply incredible. I would like to know in webrtc is it possible to add support to H265 in WS

aler9:
@brunobaraobr streaming H265 to the browser will be possible through Low-Latency HLS, please follow #469

Unfortunately WebRTC doesn't support H265 except when it's used inside macOS Safari, therefore at the moment it wouldn't be much useful to implement it.

PR #1323, opened 2022-12-22
(webrtc) added support for ice mux tcp and nat1to1ips

andrew-ld:
With this pull request introduce the following features:
- TCP mux configuration support for webrtc.
- Support for configuring a set of fixed public ip instead of webwrtc automatic discovery.

Configuration will have some new parameters:
- webrtcICETCPMuxEnable (boolean)
- webrtcICETCPMuxAddress (string)
- webrtcICEHostNAT1To1IPs (list of string)

Using webrtc ice tcp mux allows your software to work properly on network infrastructures (server side) where you have more stringent limitations, in our case we needed it to work through a single tcp port, for example.

We need NAT1To1IPs because the software runs inside a container where it is exposed to the internet via a virtual network interface, without specifying NAT1To1IPs to webrtc the candidate will get an invalid ip.


dottgonzo:
please allow to use this server with this kind of features or it will be quite impossible to use the webrtc stream inside containerized environments. This features enable useful functionalities that are right now embedded in software like wowza.

andrew-ld:
I applied codecov's suggestions, now it should be fine

aler9:
Hello,

* add the new parameters to `rtsp-simple-server.yml` with a brief description
* add the new parameters to `apidocs/openapi.yaml`
* add the new parameters to `closeWebrtcServer` in `core.go`, otherwise hot reloading / API will not work 
* `tcpMuxLn` is not closed when `webrtcServer` is closed - you have to save it somewhere and close it after `s.ln.Close()`

thanks

andrew-ld:
@aler9 I should have done everything

aler9:
This has been merged with some changes, thanks for providing the code.

I'm interested too it in the possibility of running the server inside a container without `--network=host`, i'll add to the README a section that describes how to use these new parameters in order to achieve so.

Furthermore, i'll also dig into `SetICEUDPMux()`, and if it works, i'll add a parameter for that too.

aler9:
Added in v0.21.1

PR #1341, opened 2022-12-29
basic auth credentials for api

andrew-ld:


aler9:
Hello there are two possible ways to protect the API listener:

* not protecting the API at all, add a reverse proxy before it and let the reverse proxy handle the authentication.
* protecting the API, but i think that basic auth will not be enough for a lot of users that will ask for more advanced authentication techniques (separate credentials / JWT / external authentication)

We can certainly work towards the second approach, but keep in mind that handling security directly is not a "best practice" approach since there's a high risk of causing security issues. In my work i always try to avoid developing security components directly, i generally use a series of rugged security components (identity servers, gateways, WFAs, etc) and i put them before my services.

Anyway, we can certainly add a basic security layer to the server too, but i think that basic authentication will not be enough.

andrew-ld:
having a simple level of protection like basic auth would be useful, it would avoid adding additional software such as reverse proxy and its maintenance will still remain almost zero cost since everything is implemented by gin

aler9:
@eravellaSC i'm working on an authentication system that will support explicit credentials, external authentication or JWTs along the entire server, including paths and API.

In the meanwhile, in order to protect the API, you can use a reverse proxy.

t3therdev:
Hi @aler9 any update on this.

andrew-ld:
the implementation of jwt seems like something that is very time consuming, many people simply want to block requests to the api via basic auth, this solution however remains very simple to maintain and solves many of the cases

aler9:
replaced by #3081

Issue #1416, opened 2023-01-24
Support RPi Camera Autofocus


azsde:
## Describe the feature

Some camera modules (including the newest V3 module from Raspberry Pi) do support different autofocus mode, it would be a great addition to support them as parameters.

azsde:
I implemented it myself as it was not too difficult.

As stated in the PR, the only parameter I did not include is the AutoFocus Window as I did not fully understand how it was supposed to work.

The documentation specifies a single rectangle but the libcamera defines multiple windows ?

`extern const Control<Span<const Rectangle>> AfWindows;`

aler9:
Fixed by #1417

PR #1417, opened 2023-01-24
[DEV #1416] Add support for libcamera Autofocus parameters

azsde:
I added most of the parameters as per the official documentation: https://www.raspberrypi.com/documentation/computers/camera_software.html

The only parameter I did not include is the AutoFocus Window, I did not fully understand how it was supposed to work, the documentation specifies a single rectangle but the libcamera defines multiple windows ?

`extern const Control<Span<const Rectangle>> AfWindows;`


azsde:
@aler9 I fixed the go formatting but I don't know how to trigger the checks again

aler9:
Hello, looks good to me, thanks for the contribution, if you also want to add the afWindows parameter take a look at libcamera-apps:

https://github.com/raspberrypi/libcamera-apps/blob/bce2788644fd25ec1470d18607bb14e1b31b9d23/core/libcamera_app.cpp#L502

window count is fixed to 1. In my opinion libcamera allows multiple windows because it has been designed to be as generic as possible, since its aim is to become the main Linux device API, but that doesn't mean that the RPI Camera supports multiple windows.

azsde:
Alright, I'll probably make another patch tomorrow to support this parameter so that all of the autofocus parameters are supported.

Amazing software by the way, I'm quite amazed by the performance and the ease of use.

azsde:
@aler9 I added the support for autofocus window parameter :)

Feel free to provide some feedback if needed.

aler9:
merged, thank you very much

azsde:
@aler9 you're welcome :-)

aler9:
added in v0.21.3

PR #1470, opened 2023-02-14
delete existing path onPathClose

ruokeqx:
When I try to connect the server on the second time, the connection restart infinitely

https://user-images.githubusercontent.com/55469170/218749217-f2cef095-8e32-40a6-9981-674f7f7767c7.mp4

I find that the paths in the `pathManager`  was not deleted in `onPathClose` , thus not creating new path when I connect on the second time. see https://github.com/aler9/rtsp-simple-server/blob/main/internal/core/path_manager.go#L262-L264


aler9:
thanks for reporting the issue, this is fixed in v0.21.4.

PR #1471, opened 2023-02-15
Add Webhook and Redis For Stream ready Status Events

usama-liaqat:
I have added Redis and Http Webhook to publish stream status on stream ready and when stream killed or not ready you other servers can subscribe and take actions based on the stream status


aler9:
Hello,

* webhooks are already supported, it's enough to use `wget` inside `runOnReady`, `runOnInit`, `runOnDemand`, `runOnReady`. Furthermore, your implementation of webhooks is blocking, i.e. the server waits for a webhook to return before resuming operations, and this has caused problems in the past. Existing hooks are non-blocking and are run in a separate routine.
* The same applies to Redis events. You can already use `redis-cli` inside `runOn*`, and the existing implementation is non-blocking.

Of course your implementation is more efficient than using shell commands, but at the moment i prefer not to embed specific integrations into the server, since it would result in another slice of code to maintain and to upgrade.

usama-liaqat:
if you agree to embed my code so i will make sure to upgrade my code and fix reported errors. on priority.

PR #1554, opened 2023-03-10
Ignore muxer errors for hls

andrew-ld:
fix this issue https://github.com/aler9/rtsp-simple-server/issues/1551
workaround this issue https://github.com/aler9/rtsp-simple-server/issues/1552


aler9:
see here https://github.com/aler9/rtsp-simple-server/issues/1552#issuecomment-1466655837

PR #1567, opened 2023-03-12
hls muxer: add option to dump segments on disk (#1322)

### LINKED ISSUE #1322: Save HLS fragments and playlist in file system (HDD/SSD) for less RAM footprint
## Describe the feature

Is it possible to store video fragments on file system?
I have about 20 cameras and want to have access to their videos last 48h. IMO it is too many fragments to keep them RAM but would be easy with 1TB SSD :-)

aler9:
Fixes #1322

### DIFF
--- a/internal/core/hls_muxer.go
+++ b/internal/core/hls_muxer.go
@@ -9,6 +9,8 @@ import (
 	"io"
 	"net"
 	"net/http"
+	"os"
+	"path/filepath"
 	"sync"
 	"sync/atomic"
 	"time"
@@ -56,7 +58,6 @@ type hlsMuxerParent interface {
 }
 
 type hlsMuxer struct {
-	name                      string
 	remoteAddr                string
 	externalAuthenticationURL string
 	alwaysRemux               bool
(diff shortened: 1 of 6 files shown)


internal/core/hls_muxer.go
```diff
@@ -296,6 +298,13 @@ func (m *hlsMuxer) runInner(innerCtx context.Context, innerReady chan struct{})
 			"the stream doesn't contain any supported codec (which are currently H264, H265, MPEG4-Audio, Opus)")
 	}
 
+	var muxerDirectory string
+	if m.directory != "" {
+		muxerDirectory = filepath.Join(m.directory, m.pathName)
+		os.MkdirAll(muxerDirectory, 0o755)
+		defer os.Remove(muxerDirectory)
```
andrew-ld, line 305: it is possible to set up this function to allow chunks to be kept even if the muxer goes wrong?, which can happen since many cameras do not strictly comply with standards.

aler9: @andrew-ld we can surely setup a failback mechanism, i'll reply to your issues and PR ASAP.

aler9:
added in v0.22.0

PR #1749, opened 2023-05-01
Add RTSP, RTMP, HLS, WebRTC default path support

nonoo:
If no stream path is given by a client, then use the default path given in the config. Example:

- The server has the rtmpDefaultPath set in the config to "abcd" (without the quotation marks of course)
- The server is publishing a stream called "abcd", it can be read with the URL rtmp://serverhostname/abcd
- A client opens the URL rtmp://serverhostname/
- The server detects that the client has no path in the URL, so appends the default path "abcd", so it will be served to the client.

### DIFF
--- a/internal/conf/conf.go
+++ b/internal/conf/conf.go
@@ -216,14 +216,16 @@ type Conf struct {
 	ServerKey         string      `json:"serverKey"`
 	ServerCert        string      `json:"serverCert"`
 	AuthMethods       AuthMethods `json:"authMethods"`
+	RTSPDefaultPath   string      `json:"rtspDefaultPath"`
 
 	// RTMP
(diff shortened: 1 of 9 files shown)


aler9:
Hello, this server works by placing every different stream in a separate path. Hiding the path concept would lead to confusion from new users and to a lot of issues. Why do you need this feature?

nonoo:
The PR does not hide the path. If the user does not enter a path then
mediamtx will serve the default path set in the config.

This is useful if you want to serve only 1 path or a main path when a user
only opens the server address without a path.

aler9:
Sorry, this won't be merged (for the moment).

nonoo:
Would you merge a PR which implements path rewriting based on regexp rules?

PR #1763, opened 2023-05-04
Feature keyframe warn

aler9:


alpe12:
@aler9 

My WEBRTC streams always freezes every few minutes.
With the latest update I'm getting these errors.
`WAR [path cam] no H264 key frames received in %!v(MISSING), stream can't be decoded`
My camera records with large keyframes interval when there is little motion, to save disk space.
Could this be causing issues with WEBRTC in meadiamtx? Via RTSP on mediamtx it mostly works. It freezes only (or way more) on WEBRTC.
What to do to fix WEBRTC then?
If this doesn't explain WEBRTC freezing (and not RTSP) can you add an option to disable these warning or to configure the allowed keyframe interval?

I really don't understand why only WEBRTC freezes.
Thanks.

PR #1774, opened 2023-05-08
fix: metrics listener address sanitization

tnyeanderson:
In commit 3475762 from PR #1678 the restrictNetwork function was called twice for the metrics listening address only, which leads to 0.0.0.0 listeners not working properly for the metrics server.


tnyeanderson:
@aler9 please note that I haven't tested this out, but it makes intuitive sense based on the other changes from that commit. Seems like it was just a typo before.

aler9:
merged, thanks

aler9:
added in v0.23.0

Issue #1096, opened 2022-08-17
Add option to set the "Range" header when proxying a RTSP stream


aler9:
## Describe the feature

Add option to set the "Range" header when proxying a RTSP stream, i.e.

```yml
paths:
  server:
    source: rtsp://myserver
    rtspRange: 15.0-30.0
```

This allows to seek on-demand streams.

aler9:
Fixed by #1780

PR #1780, opened 2023-05-09
Support RTSP configurable Range header

Jordy84:
Send `Range` RTSP header when source is RTSP and configuration options `rtspRangeType` & `rtspRangeStart` are configured.

### DIFF
--- a/apidocs/openapi.yaml
+++ b/apidocs/openapi.yaml
@@ -171,13 +171,10 @@ components:
     PathConf:
       type: object
       properties:
-        # source
         source:
           type: string
-        sourceProtocol:
-          type: string
-        sourceAnyPortEnable:
-          type: boolean
+
+        # general
         sourceFingerprint:
           type: string
         sourceOnDemand:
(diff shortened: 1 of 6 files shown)


Jordy84:
In our application we need to set the Start time of a recording played back using a RTSP stream. This can be done using a Range header which is already supported in `gortsplib`. This PR makes this available in Mediamtx as well.

This is my first time working with Golang, so please be gentle ;)

Hopefully this can be added to Mediamtx so we don't have to use our own fork anymore in the future

aler9:
Hello, this is very useful and it also solves issue #1096. Let me know when the PR is ready to be merged.

Jordy84:
Thanks, I think it's ready. So please let me know if you think somethings need to change

aler9:
i did some changes:
* i rebased the PR against main
* errors inside createRangeHeader() are now handled
* tests have been added
* API docs have been updated

this has been merged, thanks!

Jordy84:
Thanks for merging this 👍

aler9:
added in v0.23.0

PR #1789, opened 2023-05-11
build(script) add darwin arm64

a-wing:


aler9:
merged, thanks

aler9:
added in v0.23.0

PR #1825, opened 2023-05-18
fix missing H264/H265 keyframe warning message

aler9:


alpe12:
@aler9 
Do the time need to be 10s (for my camera it is at least)? My camera have a really long time between keyframes to save bandwidth. So I'm getting these messages all the time.
See https://github.com/bluenviron/mediamtx/pull/1763#issuecomment-1552061119
Thanks.

PR #1841, opened 2023-05-19
Expose E-Tag header for cross-origin WHEP requests

mrlt8:
Add the `access-control-expose-headers: E-Tag` header to the `/whep` response to allow access to the `E-Tag` header when making a cross-origin request.


aler9:
i added `Accept-Patch` and `Link` too and merged, thanks.

PR #1868, opened 2023-05-27
allow using special characters in external commands (#1652)

aler9:
on Windows, when using cmd.exe or a bat file as external command.


aler9:
fixed in v0.23.4

PR #1871, opened 2023-05-28
restore sending session ID to external authentication

aler9:
this fixes a regression introduced in v0.23.0


aler9:
added in v0.23.4

PR #1879, opened 2023-05-30
Add WebRTC stream id to whep response headers

vvitkovsky:
Hello, 

We plan to use Mediamtx for streaming WebRTC video / audio from local recorders like Dahua, but during testing we have found one issue - there is no information about internal WebRTC stream id in /whep request / response flow, so we don't have any information about stream that was created and cannot use Mediamtx HTTP API to list current WebRTC connections and find a stream that we just created.

So, I have a proposal to add this WebRTC stream id to /whep response headers, that will help to set info about current created stream and then manage it using Mediamtx HTTP API.

Best regards, 
Victor


vvitkovsky:
Thank you a lot!

Best regards,
Victor

aler9:
added in v0.23.4

PR #1885, opened 2023-05-31
update documentation

aler9:


aler9:
i renamed the header from UUID and added tests. Merged, thanks!

PR #1899, opened 2023-06-01
rtmp: fix timestamp conversion from RTSP/HLS to RTMP

aler9:
this was causing moments of silence and timing errors when reading with RTMP a stream originally published with RTSP or HLS.


aler9:
added in v0.23.4

Issue #1962, opened 2023-06-18
Return the path name in API RTSP session GET request 


jakew009:
## Describe the feature

At the moment the API request for `http://localhost:9997/v2/rtspsessions/list` returns something like this:

```
{
  "pageCount": 0,
  "items": [
    {
      "id": "string",
      "created": "string",
      "remoteAddr": "string",
      "state": "idle",
      "bytesReceived": 0,
      "bytesSent": 0
    }
  ]
}
```

Would it be possible to add the name of the path the session is connected to? At the moment you have to make multiple API requests to look it up.

aler9:
Hello, this is fixed by #2022 2022

aler9:
added in v0.23.8

PR #2022, opened 2023-07-05
api: add path to RTMP connections, RTSP sessions, WebRTC sessions (#1962)

volodymyr-borodin:
Hello Alessandro,

Thanks for your great work! I have a project in mind where I need to create a webpage displaying a list of all active RTMP broadcasts, along with links to their corresponding WebRTC streams. Currently, I am using the /v2/rtmpconns/list endpoint to fetch all the RTMP streams and filter them based on their status. However, I am facing an issue as the path field is missing, and without it, I cannot construct the link to the WebRTC stream.

Hope this functionality will be useful for others too. Looking forward for your comments/remarks

### DIFF
--- a/apidocs/openapi.yaml
+++ b/apidocs/openapi.yaml
@@ -389,6 +389,8 @@ components:
         state:
           type: string
           enum: [idle, read, publish]
+        path:
+          type: string
         bytesReceived:
           type: integer
           format: int64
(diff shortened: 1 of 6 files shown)


aler9:
Thanks for providing this, this fixes #1962. I added `path` to RTSP and WebRTC sessions too, added tests and merged.

PR #2040, opened 2023-07-10
Fix: API v2/paths/get with nested path always results in 404

Jordy84:
Getting path information for nested paths always fails and gives a 404. This PR fixes this problem

### DIFF
--- a/internal/core/api.go
+++ b/internal/core/api.go
@@ -233,7 +233,7 @@ func newAPI(
 	}
 
 	group.GET("/v2/paths/list", a.onPathsList)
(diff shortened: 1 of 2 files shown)


aler9:
Thanks for spotting and fixing this bug - could you do the same for these endpoints too?

```
/v2/config/paths/add/*name
/v2/config/paths/edit/*name
/v2/config/paths/remove/*name
/v2/hlsmuxers/get/:name
```

jordybsecurite:
Only the /v2/hlsmuxers/get endpoint was affected, pushed the fix for that as well

aler9:
merged, thanks!

PR #2081, opened 2023-07-19
fix a compile error with recent libcamera

hideaki-t:
not sure if this is backward compatible with older version of libcamera, but there was an API change in recnet libcamera

see https://github.com/raspberrypi/libcamera/commit/26a4b83d1b9443795dbcb21fa6c23117f113216a


aler9:
Hello, yes, this is backward-compatible. Libcamera authors are always creative and every two months they make the library incompatible with everything that has been built until them. What can i say, since Raspberry Pi authors forced us all to switch to libcamera, there are no alternatives.

hideaki-t:
ah I see... understood. Thanks.

PR #2144, opened 2023-08-03
Create GitHub action to build binaries

DanTheMan827:
This uses workflow_dispatch, so it won't run unless triggered manually.

It could be useful for people who fork the repo.


aler9:
Hello, this is a good idea, merged, thanks.

PR #2145, opened 2023-08-03
Add video player options via query string.

DanTheMan827:
This adds support for specifying controls, autoplay, muted, and playsinline via query strings.

Example:

`?controls=0&muted=0`

The default values remain unchanged, but this allows the user to specify them.

### DIFF
--- a/internal/core/hls_index.html
+++ b/internal/core/hls_index.html
@@ -19,15 +19,11 @@ html, body {
 </head>
 <body>
 
-<video id="video" muted controls autoplay playsinline></video>
-
 <script src="https://cdn.jsdelivr.net/npm/hls.js@1.4.10"></script>
 
 <script>
 
-const create = () => {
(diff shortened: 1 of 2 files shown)


aler9:
Hello, this is another good idea that improves the flexibility of web clients without touching the configuration or adding complexity to the UI. I need to check this for XSS and the i'll merge it.

DanTheMan827:
@aler9 I've also noticed that on occasion, the WebRTC player will freeze, and need to be paused and resumed to get it back to normal.

adding this code somewhere should fix it, but I don't know how you'd want to handle it.
```javascript
setInterval(() => {
    if (video.paused) {
        console.log("Video paused, resuming.");
        video.pause();
        video.play();
    }
}, restartPause);
```

aler9:
Tested and merged, thanks!

PR #2207, opened 2023-08-15
Add Webhook For Stream Events on Publish Start and On Publish Stop

usama-liaqat:
Webhook for stream publish event if stream started publishing or stopped.


aler9:
This won't be merged for reasons that were already explained to you.

PR #2232, opened 2023-08-22
allow RTMP streaming with codecid=av01 or hvc1

ph0b:
Prior to this change, when trying to stream AV1 over enhanced RTMP using XSplit Broadcaster, the server was refusing the content with "unsupported video codec: av01" message.


aler9:

aler9:
added in MediaMTX v1.0.1

PR #2303, opened 2023-09-06
add /whep suffix to documentation of whep endpoints

tysongf:


aler9:
Hello the path format "/pathname/whep" is used in this server but is not part of a standard and is not implemented by any other server. Therefore, the WHEP URL format indicated in the configuration must remain generic.

tysongf:
It might be worth mentioning somewhere in these areas of of the docs that mediamtx does use the /whep suffix for its endpoints. Or make the endpoints like whep://domain/mystream without the /whep suffix. This could have saved me some time in troubleshooting because I didn't realize mediamtx has /whep suffix for its WebRTC endpoints. It is in the docs, just not where I expected it.

aler9:
It would be fantastic to be able to remove the /whep prefix but unfortunately it's currently not technically possible, since WHIP and WHEP have the same exact handshake, so it would be impossible to distinguish between them

PR #2335, opened 2023-09-13
Configurable hlsDurationRequiredPartCount

Jordy84:
Gohlslib PR ["Start streaming parts before first completed segment in HLS Low Latency"](https://github.com/bluenviron/gohlslib/pull/92) supports configuring the required part count to be able to determine the duration. This PR makes it configurable using mediamtx.yml


aler9:
closed since the gohlslib patch is still stuck for reasons explained here: https://github.com/bluenviron/gohlslib/pull/92#issuecomment-1728445795

PR #2355, opened 2023-09-16
add runOnDisconnect, runOnNotReady, runOnUnread (#1464)

### LINKED ISSUE #1464: runOn__Stop - Add extra command to send on stop?
## Describe the feature

Could we get a separate runOn__Stop option that will run a command on stop? 

I know this can be achieved by waiting for the SIGINT signal on the regular **runOn** commands, but waiting for the signal requires an extra process to be running in the background just waiting there. 

It would be great if we could have an extra option that could trigger a command when the SIGINT would normally be sent. This would help run some cleanup scripts when something stop.

```
RunOnDemand
RunOnDemandStop

RunOnReady
RunOnReadyStop

RunOnRead
RunOnReadStop

runOnConnect
runOnConnectStop
```

aler9:
Fixes #1464


aler9:
added in v1.1.0

PR #2385, opened 2023-09-20
Support SRT encryption passphrases on configured paths

rmcnew:
Add support for SRT encryption passphrases on configured MediaMTX paths.

SRT supports encryption to protect payload data.  SRT Passphrases must be strings between 10 and 79 characters long.  Setting a SRT passphrase means that a publisher or reader using SRT MUST use the passphrase to publish or read.

Note that setting SRT passphrases does not prevent unencrypted reading from the path using another protocol!


aler9:
Hello, this is a useful feature but this PR needs a couple of improvements:

* move checks from `checkSrtPassphrase()` to `internal/conf/path.go`
* avoid using `getPathConfs()` and `pathConfs[pathName]`, `pathName` may not correspond to a configuration with the same name, since configurations can use regular expressions as names; use `res.path.conf`, that is available after `addReader()` and `addPublisher`
* fix lint errors

thanks

rmcnew:
Thank you for the feedback.  Making changes to fix those items.

rmcnew:
Updates made based on feedback and lint output.

rmcnew:
Added unit tests for CheckSrtPassphrase.  Use safeConf to prevent race conditions.

aler9:
i added some improvements:

* update API docs
* split configuration checks from connection checks
* add tests
* rename publishSRTPassphrase into srtPublishPassphrase, readSRTPassphrase into srtReadPassphrase
* remove redundant alias

and merged. Thanks!

PR #2460, opened 2023-10-05
Supports webrtc interface filtering on server

PieterFabry:
For security and ICE optimization purposes, it would be good to add the ability to restrict ICE candidates on the server based on the network interfaces available. This is a common feature for WebRTC servers.

Usage:

Added a  webrtcICEInterfaces interface to the mediamtx.yml under webrtc.
webrtcICEInterfaces takes in a string array of the network interfaces available on the server for ICE candidate resolution to happen.

eg:
webrtcICEInterfaces: [ "eno1", "enp2s0"]

The string array is then applied on the SetInterfaceFilter method in pion.
//https://pkg.go.dev/github.com/pion/webrtc/v3#SettingEngine.SetInterfaceFilter


aler9:
Hello, yes, normally the server sends the list of all its local IPs to clients, and this option allows to filter this list.

Change the code in order to fix CI errors and this will get merged. Thanks.

aler9:
merged, thanks!

aler9:
added in v1.2.0

PR #2486, opened 2023-10-10
Don't allow the webrtc player to pause

DanTheMan827:
I don't know why, maybe it's my specific sources, but there are times where the player will pause without throwing an error, this ensures that the player will always be in the playing state.

The player controls show that the video is still playing, but the video object says that the video is paused, this is why I `pause()` and then `play()`.

### DIFF
--- a/internal/core/webrtc_read_index.html
+++ b/internal/core/webrtc_read_index.html
@@ -375,6 +375,14 @@ const initVideoElement = (callback, container) => {
 		setVideoAttributes(video);
 		container.append(video);
 		callback(video);
+
+		setInterval(() => {
+			if (video.paused) {
+				console.log("Video paused, resuming.");
+				video.pause();
+				video.play();
+			}
+		}, restartPause);
 	};
 };
 


aler9:
Hello, this patch adds a background job that increases CPU consumption, thus, before merging, i'd like to understand why the stream gets paused. Can you post a server log that shows the moment in which the stream is paused? do you have a file or a sample stream that can be used to replicate the issue?

internal/core/webrtc_read_index.html
```diff
@@ -375,6 +375,14 @@
 		setVideoAttributes(video);
 		container.append(video);
 		callback(video);
+
+		setInterval(() => {
+			if (video.paused) {
+				console.log("Video paused, resuming.");
+				video.pause();
```
rgl, line 382: interesting, do you also have to put it in pause first?

rodrigoratan:
> Hello, this patch adds a background job that increases CPU consumption, thus, before merging, i'd like to understand why the stream gets paused. Can you post a server log that shows the moment in which the stream is paused? do you have a file or a sample stream that can be used to replicate the issue?

you mean the setInterval clientside code?

rodrigoratan:
> I don't know why, maybe it's my specific sources, but there are times where the player will pause without throwing an error, this ensures that the player will always be in the playing state.
> 
> The player controls show that the video is still playing, but the video object says that the video is paused, this is why I `pause()` and then `play()`.

here my streams also pause but the icon change to the play icon (meaning it`s paused)

aler9:
this can't be merged without further details.

PR #2453, opened 2023-10-03
webrtc: expose the location header

rgl:
The [suggested WHEP reference client](https://github.com/bluenviron/mediamtx/issues/2424#issuecomment-1736839160) is not able to access the `Location` header at:

https://github.com/medooze/whip-whep-js/blob/v1.0.3/whep.js#L86-L87

This PR allows the browser to expose that header to the javascript code.

PS: It [also needs to use the `DELETE` method](https://github.com/medooze/whip-whep-js/blob/v1.0.3/whep.js#L481-L485). should I add it to this PR too? or another PR?

Though, this is still not enough to get that client working, as it fails with:

```
2023/10/03 20:52:30 DEB [WebRTC] [conn [::1]:44266] PATCH /camera_sub/whep
2023/10/03 20:52:30 DEB [WebRTC] [conn [::1]:44266] [c->s] PATCH /camera_sub/whep HTTP/1.1
Host: localhost:8889
Accept: */*
Accept-Encoding: gzip, deflate, br
Accept-Language: en-US,en;q=0.9,pt-PT;q=0.8,pt;q=0.7
Connection: keep-alive
Content-Length: 105
Content-Type: application/trickle-ice-sdpfrag
Origin: http://localhost:8080
Referer: http://localhost:8080/
Sec-Ch-Ua: "Google Chrome";v="117", "Not;A=Brand";v="8", "Chromium";v="117"
Sec-Ch-Ua-Mobile: ?0
Sec-Ch-Ua-Platform: "Linux"
Sec-Fetch-Dest: empty
Sec-Fetch-Mode: cors
Sec-Fetch-Site: same-site
User-Agent: Mozilla/5.0 (X11; Linux x86_64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/117.0.0.0 Safari/537.36

a=ice-ufrag:w08e
a=ice-pwd:6j8Xfsi7cdr8nYnyvEuoT/vT
m=video 9 RTP/AVP 0
a=mid:0
a=end-of-candidates

2023/10/03 20:52:30 DEB [WebRTC] [conn [::1]:44266] [s->c] HTTP/1.1 400 Bad Request
Access-Control-Allow-Credentials: true
Access-Control-Allow-Origin: *
Server: mediamtx


2023/10/03 20:52:36 DEB [WebRTC] [session 3c3ca543] peer connection state: disconnected
2023/10/03 20:52:36 DEB [WebRTC] [session 3c3ca543] peer connection state: closed
[...]

PR #2507, opened 2023-10-15
webrtc: implement the whep http delete method

rgl:
this actually does nothing more than returning OK to satisfy the [whep.js `DELETE` method](https://github.com/medooze/whip-whep-js/blob/v1.0.3/whep.js#L481-L485) as mentioned in https://github.com/bluenviron/mediamtx/pull/2453.

### DIFF
--- a/internal/core/api_test.go
+++ b/internal/core/api_test.go
@@ -809,7 +809,7 @@ func TestAPIProtocolList(t *testing.T) {
 				defer source.Close()
 
 				c := newWebRTCTestClient(t, hc, "http://localhost:8889/mypath/whep", false)
-				defer c.close()
+				defer c.close(t, true)
 
 				time.Sleep(500 * time.Millisecond)
 
@@ -1095,7 +1095,7 @@ func TestAPIProtocolGet(t *testing.T) {
 				defer source.Close()
 
 				c := newWebRTCTestClient(t, hc, "http://localhost:8889/mypath/whep", false)
-				defer c.close()
+				defer c.close(t, true)
 
 				time.Sleep(500 * time.Millisecond)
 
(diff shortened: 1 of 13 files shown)


aler9:
Hello, i'd like to avoid merging workarounds and instead merge complete implementations. The DELETE method in WHEP has the role of terminating a session, by using a session identifier, that is set by the server through POST responses, through the Location header.

The server currently doesn't put the session identifier (secret) into the Location header, but puts it into the ETag header, which is compatible with PATCH but not with DELETE. Therefore, a complete TODO list for implementing this feature is:

1. move the session secret from the `Etag` to the `Location` header, after the path name (`/mypath/whip/secret`) and adjust current endpoints (POST and PATCH) in order to support this change. The `ETag` header on the other hand has to be set to `*` according to the specification.
2. Add the `DELETE` endpoint and use it to terminate the WebRTC session, by using the secret.

Examples of POST and PATCH requests that make use of the Location and ETag header are here: https://datatracker.ietf.org/doc/draft-ietf-wish-whip/

rgl:
@aler9 please check it now.

I need your help to understand how to actually implement the `DELETE` method.

I also think that `secret` should be renamed to something else. Maybe `sessionId`? OTOH, there is an `Id` header flowing around, which I'm not sure why its needed, and maybe it not needed anymore?

aler9:
> I also think that secret should be renamed to something else. Maybe sessionId? OTOH, there is an Id header flowing around, which I'm not sure why its needed, and maybe it not needed anymore?

This is a good question - the difference between secret and ID is that the latter is printed in logs. If we store the secret in logs or use the same variable for secret and ID, we would create a security issue, since anyone able to access logs would be able to hijack sessions.

rgl:
> > I also think that secret should be renamed to something else. Maybe sessionId? OTOH, there is an Id header flowing around, which I'm not sure why its needed, and maybe it not needed anymore?
> 
> This is a good question - the difference between secret and ID is that the latter is printed in logs. If we store the secret in logs or use the same variable for secret and ID, we would create a security issue, since anyone able to access logs would be able to hijack sessions.

With this pr, the secret is now effectively in the url, which we should assume will end up in the logs. Maybe the url should use the ID instead?

aler9:
i finished the implementation and merged, thanks!

aler9:
added in v1.2.1

PR #2582, opened 2023-10-27
api: return 500 instead of 404 in case of non-existing items

aler9:
This is needed to differentiate between calls to non-existent andpoints and calls to valid endpoints, but with an ID pointing to a non-existing item, and to return JSON in the latter case.

### DIFF
--- a/apidocs/openapi.yaml
+++ b/apidocs/openapi.yaml
@@ -634,7 +634,7 @@ paths:
         '400':
           description: invalid request.
         '500':
-          description: internal server error.
+          description: server error.
 
   /v3/config/global/patch:
     patch:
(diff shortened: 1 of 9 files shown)


tbnguyen1407:
Hello, just a suggestion to intead use

- 501: for unknown/unmapped route (where we are not aware of the requested resource)
- 404: for non-existent resources (where we are aware of the resource but cannot find it)

Since 404 is very common for standard CRUD APIs, using 500 code for it seems a bit weird and we must deserialize the body and check for a hardcoded string to know if it is not found or other error.

aler9:
@tbnguyen1407 i fixed it in #2920

Issue #2167, opened 2023-08-07
Allow to set video resolution when publishing with WebRTC / browser


aler9:
## Describe the feature

Discussed in #2156

aler9:
fixed by #2584

Issue #1837, opened 2023-05-19
Allow publishing  through WebRTC and browser with pre-set parameters


iesv:
## Describe the feature

Description
0.23.1 Browser push streams can add parameters, for example
http://localhost:8889/mystream/publish?vedio=screen&acodec=h264

aler9:
fixed by #2584

PR #2584, opened 2023-10-27
webrtc: add options (width, height, framerate, audio capture) and store parameters in URL (#1837) (#2167)

Lukinoh:
Ciao,

For my personal needs, I wanted to add the audio capture feature, but I ended up adding other features, namely (#1837) (#2167).

I added these new options:
 - video width
 - video height
 - video ideal framerate
 - audio capture

And I implemented the feature to store the options in the URL.

I also did a bit of refactoring:
- Improve alignment of the form
- Reduce redundancy of `document.getElementById` for the different inputs in the form
- Use only once `new URL(window.location.href)`

### DIFF
--- a/internal/core/webrtc_publish_index.html
+++ b/internal/core/webrtc_publish_index.html
@@ -26,21 +26,26 @@ body {
     justify-content: center;
     min-height: 200px;
     padding: 10px;
-}
-#device {
     flex-direction: column;
 }
-#device > div {
-    margin: 10px 0;
+
+#device {
+    display: grid;
+    grid-template-rows: repeat(6, 1fr);
+    grid-auto-flow: column;
+    gap: 10px 20px
+}
+
+.item {
     display: flex;
     gap: 20px;
-    justify-content: center;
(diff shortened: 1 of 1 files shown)


aler9:
Hello, i've tested the PR and it performs its job. Alignment, linting and tabs are not perfect but that's not a problem, it will be handled in the future when a JS linter (and maybe a JS compiler) will be added to the building process.

For clarity, remove the "capture audio" switch and edit the page in order to always try to capture audio when sharing the screen. The label "capture audio" can cause confusion, the feature is available only in some browsers and, in case it's available, there's already a big switch in the screen capture dialog that opens shortly after.

Lukinoh:
Thank you for the feedback.

I improved a bit the alignment of the (visual) form.

Concerning the audio capture, I totally agree with you.
My first plan was to simply set it to `true`, but as it was explicitely set to `false`, I was unsure if it was voluntarily done as such or not. So I added the toggle.

aler9:
merged, thanks!

PR #2631, opened 2023-11-02
HLS.js usage: bugfix, stability improvements and resilience improvement

rse:
I propose the following changes to the HLS.js:
- BUGFIX: pass "video" element to create() on restart triggering fatal-error
- STABILITY: await MEDIA_ATTACHED event before performing "loadSource" on HLS
- STABILITY: await MANIFEST_PARSED event before performing "play" on video element
- RESILIENCE: on "MEDIA_ERROR" event perform "recoverMediaError" on HLS

### DIFF
--- a/internal/core/hls_index.html
+++ b/internal/core/hls_index.html
@@ -33,17 +33,21 @@ const create = (video) => {
 		});
 
 		hls.on(Hls.Events.ERROR, (evt, data) => {
-			if (data.fatal) {
+			if (data.type === Hls.ErrorTypes.MEDIA_ERROR)
+				hls.recoverMediaError();
+			else if (data.fatal) {
 				hls.destroy();
-
-				setTimeout(create, 2000);
+				setTimeout(() => create(video), 2000);
 			}
 		});
 
-		hls.loadSource('index.m3u8' + window.location.search);
+		hls.on(Hls.Events.MEDIA_ATTACHED, () => {
+			hls.loadSource('index.m3u8' + window.location.search);
+		});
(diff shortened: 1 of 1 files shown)


internal/core/hls_index.html
```diff
@@ -33,18 +33,22 @@
...
+				setTimeout(() => create(video), 2000);
 			}
 		});
 
-		hls.loadSource('index.m3u8' + window.location.search);
+		hls.on(Hls.Events.MEDIA_ATTACHED, () => {
+			hls.loadSource('index.m3u8' + window.location.search);
+			hls.on(Hls.Events.MANIFEST_PARSED, () => {
+				video.play();
+			});
```
aler9, line 48: move `hls.on(Hls.Events.MANIFEST_PARSED)` outside `hls.on(Hls.Events.MEDIA_ATTACHED)`, otherwise this hook will be setupped repeatedly every time the `MEDIA_ATTACHED` event is triggered.

aler9 (changes requested):
Hello, thanks for this contribution, just edit the lines i commented and this will be merged.

internal/core/hls_index.html
rse: Yes, in theory this can happen. In practice, as you just once call "attachMedia", the callback on "MANIFEST_PARSED" will actually happen just once. Nevertheless, I moved it one level out now.

rse:
Code was now adjusted.

aler9 (approved):

aler9:
merged, thanks

PR #2638, opened 2023-11-03
webrtc: restore compatibility when running behind a proxy, in a subpath

aler9:


LeRatierBretonnien:
Hi.
Using v1.3.0, and I am not sure if it has been fixed.
Proxying with apache in a subfolder does not work : error 404 is returned. Direct access to the WebRTC stream thru the 8889 port is working.
Any hint/idea ?
Thanks.

aler9:
@LeRatierBretonnien this is not the way to report bugs. Open an issue and provide details on how to replicate.

PR #2656, opened 2023-11-06
Add query as optional block for SRT publish and read

bcherniakh:
Illustrates solution 1 for #2654


apedorenkoS (approved):

aler9:
Hello, yes, this is a good way to provide the query feature to the SRT protocol too. Merged, thanks.

Issue #2491, opened 2023-10-11
HTTPS for API endpoints


eravellaSC:
## Describe the feature

Would it be possible to add encryption for the API endpoints?

PR #2658, opened 2023-11-07
feat: added TLS encryption support to API calls (HTTPS)

eravellaSC:
Solving #2491, I added support for TLS-encryption for the HTTP API calls. The gin server was already set up to do so, the only thing missing were the conf paths.

### DIFF
--- a/internal/conf/conf.go
+++ b/internal/conf/conf.go
@@ -92,6 +92,9 @@ type Conf struct {
 	ExternalAuthenticationURL string          `json:"externalAuthenticationURL"`
 	API                       bool            `json:"api"`
 	APIAddress                string          `json:"apiAddress"`
+	APIEncryption             bool            `json:"apiEncryption"`
+	APIServerKey              string          `json:"APIServerKey"`
+	APIServerCert             string          `json:"APIServerCert"`
 	Metrics                   bool            `json:"metrics"`
 	MetricsAddress            string          `json:"metricsAddress"`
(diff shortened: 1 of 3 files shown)


eravellaSC:
I double checked the failing tests and coverage reports and they mention files and checks that I have not touched. It's a very small modification, and the codebase was already ready to accomodate this simple addon. Is the coverage slowing down the review process? I can dig deeper if it's needed.

eravellaSC:
I am sorry to bother directly, @aler9, but I'd like to know if there is interest in adding this feature. If not I can close the PR.

aler9:
Hello @eravellaSC, of course there's interest, but the problem is that currently the API is missing a strong authentication system, therefore it's meant to be used either in a closed environment or behind a reverse proxy which provides authentication and TLS termination.

If i merge this feature without providing authentication, there's risk of giving the impression that the API is safe to use. 

Furthermore, nowadays valid TLS certificates are generated by authorities that periodically query the corresponding domain through HTTP, and that would mean that some users may try to expose the API publicly. Since it's easy to scan the internet and detect running instances of MediaMTX, this would mean creating a Botnet.

Anyway, the authentication issue will be solved soon.

eravellaSC:
@aler9 Thanks for the update, I didn't think of that in that way. I'll wait for the auth to be merged then.

aler9:
replaced by #3280

PR #2774, opened 2023-12-02
support unix timestamp in filenames of recorded segments

neerajkanhere:
While parsing %s (unix timestamp) takes precedence over Y-%m-%d_%H-%M-%S


aler9:
i added tests, updated documentation and merged, thanks!

PR #2828, opened 2023-12-20
RunOn.. commands - Propagate errors in restart configurations

jsbohnert:
The RunOn{Init,Ready, etc} commands call back with a generic 'exited with code 0' message if the command is specified with its relevant 'Restart' option set to 'yes'. eg

`runOnInit command exited: command exited with code 0`

This PR allows underlying `err` result to actually be seen in logs, eg

`runOnInit command exited: exec: "ffmpeg": executable file not found in %PATH%`

### DIFF
--- a/internal/externalcmd/cmd.go
+++ b/internal/externalcmd/cmd.go
@@ -87,7 +87,6 @@ func (e *Cmd) run() {
 			return
 		}
 
-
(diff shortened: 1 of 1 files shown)


aler9:
i applied `gofmt` and merged, thanks!

Issue #1681, opened 2023-04-11
Database save option


pikachu937:
## Describe the feature

Add the ability to save metadata to the pgsql database for saved files:

> file name
> the path to the file
> date of creation
> record end date
> size
> duration
> the name of the stream that owns the file

Issue #998, opened 2022-06-21
Improve Kubernetes / scalability support


aler9:
## Describe the feature

Create an operator / Helm chart that allows users to instantiate the server with native scalability support

aler9:
TODO:

1. Allow multiple server replicas to share the same internal state, by using an external database (probably etcd)
2. Distribute paths among available replicas
3. If a replica receives a publish request that doesn't belong to it, proxy the request to the correspondent replica
4. If a replica receives a read request that doesn't belong to it, contact the replica that has the correspondent stream and proxy that stream
5. Create an Helm chart that allows users to instantiate and scale the server with no effort.

pikachu937:
Support it

johnny-ark:
is it already done or not yet started?

ethan-riley:
any news on this?

docimin:
Still waiting :/

ericsouza:
This feature would be amazing. I think it's not just the scaling problem but also being able to upgrade running MediaMTX  server without having to disconnect all clients (starting a new instance, moving all connections to it and then removing the old one)

rafiuske:
Hello,

Is there a schedule for which release this feature will be launched?

j4y-v:
Hey everyone 👋

I’ve been thinking about how MediaMTX could scale on Kubernetes before native multi-replica support is available.

The idea is to use regex paths to handle many cameras with a single rule (like ~^cam_(.+)$), and a runOnDemand ffmpeg command — or a custom command that, given a camera ID, looks up a database to fetch the full RTSP URL with credentials — that starts only when a stream is requested, then publishes it back to the same path.

On the Kubernetes side, the Ingress is made sticky by path, so all viewers of the same camera hit the same pod. That way, there’s only one ffmpeg transcode per camera, even with multiple viewers.

Conceptually, it seems to work fine (on-demand startup, per-camera isolation), but I haven’t really tested it at scale yet.

Do you think this kind of setup could make sense until native multi-replica support lands? Or am I overlooking something important?

spenpw:
> On the Kubernetes side, the Ingress is made sticky by path, so all viewers of the same camera hit the same pod. That way, there’s only one ffmpeg transcode per camera, even with multiple viewers.


[...]

PR #2851, opened 2023-12-28
Add database and record audio parametrs

major1ink:
Added functionality:
Entering information about saved files into the database (PostgreSQL).
Ability to disable audio in saved files.

The configuration file has the ability to disable/enable work with the database.

Algorithm:
When a new file is created, information about the file is entered into the database. Later, when a file segment is closed, the file information is updated in the database.
If, after creating the file, the information could not be entered into the database, the created file is deleted.
This functionality is implemented for fmp4 and mpegts.

With the recordAudio: no parameter, audio format packets will be skipped to be saved to a file, but the audio stream will not be changed in the broadcast itself.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -44,6 +44,10 @@ require (
 	github.com/go-playground/universal-translator v0.18.1 // indirect
 	github.com/go-playground/validator/v10 v10.14.0 // indirect
 	github.com/goccy/go-json v0.10.2 // indirect
+	github.com/jackc/pgpassfile v1.0.0 // indirect
+	github.com/jackc/pgservicefile v0.0.0-20221227161230-091c0ba34f0a // indirect
+	github.com/jackc/pgx/v5 v5.5.1 // indirect
+	github.com/jackc/puddle/v2 v2.2.1 // indirect
 	github.com/json-iterator/go v1.1.12 // indirect
 	github.com/klauspost/cpuid/v2 v2.2.4 // indirect
 	github.com/leodido/go-urn v1.2.4 // indirect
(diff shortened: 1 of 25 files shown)


pikachu937:
Please accept

aler9:
Hello, binding the server to PostgreSQL in this way has several disadvantages:

* this adds a mandatory, additional layer (the database) that is not necessary and unwanted by most users, in particular by the ones with limited computational capabilities
* this binds the server with PostgreSQL only, which is not ideal for all scenarios
* this delegates the state of a single instance to a database in an unidirectional way, making scaling impossible and greatly complicating the implementation of #998
* this adds an additional dependency that increases maintenance efforts, executable size and building time
* delegation of state to the database is not complete, but involves recordings only, thus a much bigger effort would be required in order to port all existing features to the database
* all the features added in this PR can already be implemented by using hooks, with no particular overhead

If you want to build a solution that involves PostgreSQL and video-only recording, my advice is to use hooks in order to populate your database of choice and to generate video-only streams.

pikachu937:
Greetings.
>> in this way has several disadvantages

Once you also talked about the built-in recording function, but later you implemented it. 

>> this adds a mandatory, additional layer (the database) that is not necessary and unwanted by most users, in particular by the ones with limited computational capabilities

Using the database is optional, so users with limited computing capabilities will not lose anything; you can work with or without the database enabled.

>> this binds the server with PostgreSQL only, which is not ideal for all scenarios

Regarding the connection only with Postgresql, no one forbids adding support for other databases.

>> this delegates the state of a single instance to a database in an unidirectional way

What else would you like to delegate to the database, what functions?

>>scaling impossible and greatly complicating the implementation of #998

This functionality will not affect the scaling in any way, since recording in any case occurs on the node on which the stream is published.

>> all the features added in this PR can already be implemented by using hooks, with no particular overhead

Using hooks to work with a database is not always convenient, so a request for extending functionality #1681 was created, which you approved and marked as an improvement, which was implemented in this commit.

moenodedev:
You could store all your settings in a database then make a script that writes the yaml file and restarts the server. Or the script could perform API calls.

PR #2876, opened 2024-01-05
Log error instead of closing stream when we get a B-frame

RogerDavenport:
On long-running streams we often see a single repeat PTS from our RTSP camera, causing the WebRTC stream to close.  This patch logs it as an error instead.


aler9:
Hello, while this PR may be able to solve the issue with your specific camera, it also allows any H264 stream with B-frames to be read to WebRTC, filtering out its B-frames, resulting in a corrupted stream.

Please open an issue instead and provide a network dump in order to find out whether we can filter out the access unit with the repeated PTS in a previous step.

PR #2882, opened 2024-01-06
Ignore buggy Authentication on Grandstream SETUP message using skipAuth

RogerHardiman:
**NOTE. This is an alternative Pull Request to User-Agent checks and workarounds in gortslib.
We need either this patch, or the gortsplib patch**

Ignore buggy Authentication on Grandstream SETUP message using skipAuth and User-Agent

The Grandstream SETUP message on the GXV3500 has two bugs.

Bug 1: MediaMTX is in Basic Auth mode.
Grandstream send a Digest Authorization Header for SETUP even though MediaMTX has not supplied a Digest Realm or Nonce. This invalid Header needs to be ignored.

SETUP rtsp://192.168.26.61:8554/stream1/trackID=0 RTSP/1.0 Authorization: Digest username="admin", realm="", nonce="", uri="rtsp://192.168.26.61:8554/stream1", response="9b0161bf04d518c1c278f6f575568394"

Bug 2: MediaMTX is in Digest Auth mode.
Grandstream have a Digest Realm and Nonce but send a SETUP message with an Authorization Header where
  a) the uri does not match the SETUP message URI
  b) the response value does not match a computed hash for either the
     Message URI or the Digest URI so seems to be a corrupt value

  For info, this header is with username admin, password admin

SETUP rtsp://192.168.26.61:8554/stream1/trackID=0 RTSP/1.0 Authorization: Digest username="admin", realm="IPCAM", nonce="7011b1979fcb236596a729d46cb39a8e", uri="rtsp://192.168.26.61:8554/stream1", response="af70e5ab066e157ec9f44bdd82713d8d"

It was noted that MediaMTX does not check authentication for PLAY messages. So the workaroud skips authentication for Grandstream SETUP messages.

### DIFF
--- a/internal/servers/rtsp/session.go
+++ b/internal/servers/rtsp/session.go
@@ -5,6 +5,7 @@ import (
 	"errors"
 	"fmt"
 	"net"
+	"strings"
 	"sync"
 	"time"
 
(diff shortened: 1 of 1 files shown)


aler9:
Hello, of course this cannot be merged since it allows anyone to bypass authentication by changing the user agent, causing a huge security issue.

PR #2901, opened 2024-01-11
Endpoints for managing recordings

eravellaSC:
This is a WIP pull request (I am still working on the whole feature) that I am posting to ask if it makes sense to have this API extension. The aim here would be having some interface to manage the recorded streams via the REST API, allowing operations as 
* listing
* deletion
* streaming (still need to be added, I am trying to figure out how to stream from file or if it is easier to just create a stream and run an FFMPEG script)

Does this feature make sense?

### DIFF
--- a/apidocs/openapi.yaml
+++ b/apidocs/openapi.yaml
@@ -1819,7 +1819,7 @@ paths:
               schema:
                 $ref: '#/components/schemas/Error'
 
-  /v3/recodings/list:
+  /v3/recordings/list:
     get:
       operationId: listRecordings
       summary: Lists all the stored recordings
--- a/apidocs/openapi.yaml
+++ b/apidocs/openapi.yaml
@@ -640,6 +640,12 @@ components:
           items:
             $ref: '#/components/schemas/WebRTCSession'
 
+    RecordingsList:
+      type: array
+      items:
+        type: string
+
+
 paths:
   /v3/config/global/get:
     get:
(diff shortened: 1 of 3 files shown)


bog:
This is something I'd be very interested in.

eravellaSC:
I added tests and refactored some code. I think that the steaming back functionality is being addressed in #2906 (and it's still achievable through the already existing API and the use of hooks). I will wait for review and if there are things that need to be adjusted I will work on them.

t3therdev:
Hi @eravellaSC playback is released in 1.5. I think your suggested API enhancements would work well also.

aler9:
sorry for not including this in 1.5 but i had no time for reviewing it. Will do ASAP.

eravellaSC:
Not a problem for me. Let me know if further work is needed / if I can help with something

apidocs/openapi.yaml
```diff
@@ -1812,3 +1818,23 @@ paths:
             application/json:
               schema:
                 $ref: '#/components/schemas/Error'
+
+  /v3/recodings/list:
```
averred, line 1822:
```suggestion
  /v3/recordings/list:
```

aler9:
Unfortunately this PR lacks some critical features, including:

* handling segments of paths that use a `recordPath` value different from the one in `defaultPaths`
* retrieving path names
* segment management independent from file paths on disk
* pagination

i've created a new patch here: #2988

aler9:
thank you for providing the idea behind the feature.

PR #2962, opened 2024-01-29
SRT Metrics

nerdbaggy:
This is the start of the SRT metrics. It should all be in there minus the prometheus endpoint, that is still a WIP.

The variable names and comments were pulled from https://github.com/datarhei/gosrt/blob/main/statistics.go since he has better than I could think of.

I do not have any of the `StatisticsInterval` metrics because that can all be derived from prometheus queries.


nerdbaggy:
All tests should pass now with the exception to the lint test.

The lint test says that the line length are too long. This is due to the comments about what each metric is for. Do you want me to remove the comment or should it just be ignored for that struct? 
https://github.com/nerdbaggy/mediamtx/blob/srt-metrics/internal/defs/api.go#L165

nerdbaggy:
Should be good to go for real this time. The tests has been fixed and the lint issue has been fixed

nerdbaggy:
Is there anything else you would like me to do
in this PR? While Codecov says that it isn’t covered, the metrics code does get tested when the Prometheus endpoint is tested.

aler9:
merged, thanks!

PR #3039, opened 2024-02-18
Add tzdata to Docker images

HouCoder:
By default, the timezone will always be UTC in the container and there's no way to change it, with tzdata package preinstalled we can specify the timezone in the container by passing the `TZ` environment variable.

### DIFF
--- a/scripts/dockerhub.mk
+++ b/scripts/dockerhub.mk
@@ -1,7 +1,8 @@
 DOCKER_REPOSITORY = bluenviron/mediamtx
 
 define DOCKERFILE_DOCKERHUB
-FROM scratch
+FROM $(ALPINE_IMAGE)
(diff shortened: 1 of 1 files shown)


aler9:
Hello, in order to change the timezone in any Docker container it's enough to mount `/etc/timezone` and `etc/localtime` from the host or from a ConfigMap. This PR adds files of all timezones to the base image, furthermore it changes the base image to Alpine Linux, two changes that increase the image size and may cause incompatibility with existing setups.

If you need a more advanced image, you can create one by using the multi-stage feature of Docker:

```
FROM bluenviron/mediamtx AS mediamtx
FROM anybaseimage
COPY --from=mediamtx /mediamtx /mediamtx.yml /
```

Issue #828, opened 2022-02-11
config: environment substitution broken for commands in rtsp-simple-server.yml


tmatth:
## Which version are you using?

\>= v0.17.15

## Which operating system are you using?

- [x] Linux amd64 standard
- [x] Linux amd64 Docker
- [ ] Linux arm64 standard
- [ ] Linux arm64 Docker
- [ ] Linux arm7 standard
- [ ] Linux arm7 Docker
- [ ] Linux arm6 standard
- [ ] Linux arm6 Docker
- [ ] Windows amd64 standard
- [ ] Windows amd64 Docker (WSL backend)
- [ ] macOS amd64 standard
- [ ] macOS amd64 Docker
- [ ] Other (please describe)

## Describe the issue

Since 3a81728096bb4a6fda14be3926f9f8cdaf13c612, arbitrary environment variables will not be replaced in the `runOnDemand` command, e.g. if you had something like
```
runOnDemand: rtsp-command rtsp://${USER_DEFINED_VARIABLE}
```
where `${USER_DEFINED_VARIABLE}` is defined in `rtsp-simple-server`'s environment, it's not getting substituted so the command is executed with the literal `${USER_DEFINED_VARIABLE)`.

## Describe how to replicate the issue

With this config:
```diff
diff --git a/rtsp-simple-server.yml b/rtsp-simple-server.yml
index 2afe6f4..a586d6b 100644
--- a/rtsp-simple-server.yml
+++ b/rtsp-simple-server.yml
@@ -246,7 +246,7 @@ paths:
     # * RTSP_PORT: server port
     # * G1, G2, ...: regular expression groups, if path name is
     #   a regular expression.
-    runOnReady:
+    runOnReady: echo Hey ${FOOBAR}
     # Restart the command if it exits suddenly.
     runOnReadyRestart: no
 
```

1. start the server with ```FOOBAR=foobar ./rtsp-simple-server```
2. publish with ```ffmpeg -re -f lavfi -i testsrc -c:v vp8 -crf 23 -f rtsp rtsp://127.0.0.1:8554/test```

## Did you attach the server logs?

yes:
```
2022/02/11 00:07:41 INF rtsp-simple-server v0.0.0
2022/02/11 00:07:41 INF [RTSP] listener opened on :8554 (TCP), :8000 (UDP/RTP), :8001 (UDP/RTCP)
2022/02/11 00:07:41 INF [RTMP] listener opened on :1935
[...]

PR #3044, opened 2024-02-19
Expand host env vars

dbason:
Expand environment vars from the underlying host in the command.

The specific usecase this will solve for us is publishing the hostname to an etcd store on ready which will allow naive horizontal scalability.


aler9:
`os.Expand()` is a nice catch. Merged, thanks!

tmatth:
Thanks for this, this effectively solves a long standing issue for us that we had to patch around (previously discussed here: https://github.com/bluenviron/mediamtx/issues/828)

PR #3051, opened 2024-02-21
fix compile errors with libcamera 0.2.0

hideaki-t:
there was a breaking change in libcamera v0.2.0, I made a quick fix.
see https://github.com/raspberrypi/libcamera/commit/c65e40b8480ffb5f50e01a4e6713164c7194a937


dbason:
I think there might be a race condition in the test that is failing.  It looks like it has also failed in at least one of the other open PRs.

aler9:
replaced by https://github.com/bluenviron/mediamtx-rpicamera/pull/1

Issue #3118, opened 2024-03-07
Leak of goroutines with WHEP sources


RouquinBlanc:
## Which version are you using?

Issue seen on v1.6.0 and on master after the merge of #3110 .
Issue not present AFAICS on v1.5.1

## Which operating system are you using?

- [ ] Linux amd64 standard
- [x] Linux amd64 Docker
- [ ] Linux arm64 standard
- [ ] Linux arm64 Docker
- [ ] Linux arm7 standard
- [ ] Linux arm7 Docker
- [ ] Linux arm6 standard
- [ ] Linux arm6 Docker
- [ ] Windows amd64 standard
- [ ] Windows amd64 Docker (WSL backend)
- [ ] macOS amd64 standard
- [ ] macOS amd64 Docker
- [ ] Other (please describe)

## Describe the issue

The configuration contains a mix of WHEP sources (less than 10), with 3 connecting successfully (the remote endpoints are up) and the others not.

The number of goroutines related to pion/webrtc very quickly raise to thousands (reached 11K in less than 24 hours) - see attached goroutines from pprof.

## Describe how to replicate the issue

1. start the server with some WHEP paths pointing to nothing. It should start leaking quickly

## Did you attach the server logs?

[goroutines.txt](https://github.com/bluenviron/mediamtx/files/14526614/goroutines.txt)

yes, goroutines

## Did you attach a network dump?

no

aler9:
Hello, i tried replicating the issue but in my case the number of goroutines remained constant, so there must be some specific configuration combination that triggers the leak. This is my configuration:

```yml
paths:
  nonexisting_url:
    source: whep://127.0.0.1:8889/nonexisting/whep

  nonexisting_host:
    source: whep://nonexisting:8889/nonexisting/whep

  working:
    source: whep://127.0.0.1:8889/stream/whep
```

can you provide a configuration that allows to trigger the leak?

also, can you provide a goroutine dump by using the integrated pprof server? you need to set `pprof: yes` inside the configuration and post the output of

```
go tool pprof -text http://localhost:9999/debug/pprof/goroutine
```

RouquinBlanc:
Hello, the initial report was far from enough, apologies!

In fact, trying to isolate what's happening, I end up with the following minimal config:
```
pprof: yes

paths:
    m400_dltv_whep:
       source: whep://1.2.3.4:8889/m400_dltv/whep
```

The other side is a mediamtx instance (v1.5.1) on another machine with the following relevant config:
[...]

PR #3124, opened 2024-03-09
fix: whep gathering failure leaks peer connections

RouquinBlanc:
Tentative fix for #3118

It still leaks, but less; Here is what remains leaking:

```
goroutine profile: total 128
50 @ 0x100843218 0x10083cae8 0x100876730 0x1008b8538 0x1008b9880 0x1008b9871 0x100a20cb8 0x100a32234 0x100b5ddd0 0x100ac7488 0x100ac75f0 0x100b5ed58 0x10087d084
#	0x10087672f	internal/poll.runtime_pollWait+0x9f		/opt/homebrew/opt/go/libexec/src/runtime/netpoll.go:345
#	0x1008b8537	internal/poll.(*pollDesc).wait+0x27		/opt/homebrew/opt/go/libexec/src/internal/poll/fd_poll_runtime.go:84
#	0x1008b987f	internal/poll.(*pollDesc).waitRead+0x1ff	/opt/homebrew/opt/go/libexec/src/internal/poll/fd_poll_runtime.go:89
#	0x1008b9870	internal/poll.(*FD).Read+0x1f0			/opt/homebrew/opt/go/libexec/src/internal/poll/fd_unix.go:164
#	0x100a20cb7	net.(*netFD).Read+0x27				/opt/homebrew/opt/go/libexec/src/net/fd_posix.go:55
#	0x100a32233	net.(*conn).Read+0x33				/opt/homebrew/opt/go/libexec/src/net/net.go:179
#	0x100b5ddcf	net/http.(*persistConn).Read+0x4f		/opt/homebrew/opt/go/libexec/src/net/http/transport.go:1976
#	0x100ac7487	bufio.(*Reader).fill+0xf7			/opt/homebrew/opt/go/libexec/src/bufio/bufio.go:110
#	0x100ac75ef	bufio.(*Reader).Peek+0x5f			/opt/homebrew/opt/go/libexec/src/bufio/bufio.go:148
#	0x100b5ed57	net/http.(*persistConn).readLoop+0x157		/opt/homebrew/opt/go/libexec/src/net/http/transport.go:2140

50 @ 0x100843218 0x1008566b8 0x100b60600 0x10087d084
#	0x100b605ff	net/http.(*persistConn).writeLoop+0x9f	/opt/homebrew/opt/go/libexec/src/net/http/transport.go:2443
```

### DIFF
--- a/internal/protocols/webrtc/whip_client.go
+++ b/internal/protocols/webrtc/whip_client.go
@@ -78,6 +78,7 @@ func (c *WHIPClient) Publish(
 
 	err = c.pc.SetAnswer(res.Answer)
 	if err != nil {
+		WHIPDeleteSession(context.Background(), c.HTTPClient, c.URL.String()) //nolint:errcheck
 		c.pc.Close()
 		return nil, err
 	}
@@ -91,6 +92,7 @@ outer:
 		case ca := <-c.pc.NewLocalCandidate():
 			err := WHIPPatchCandidate(ctx, c.HTTPClient, c.URL.String(), offer, res.ETag, ca)
 			if err != nil {
+				WHIPDeleteSession(context.Background(), c.HTTPClient, c.URL.String()) //nolint:errcheck
 				c.pc.Close()
 				return nil, err
 			}
@@ -101,6 +103,7 @@ outer:
 			break outer
 
 		case <-t.C:
+			WHIPDeleteSession(context.Background(), c.HTTPClient, c.URL.String()) //nolint:errcheck
 			c.pc.Close()
 			return nil, fmt.Errorf("deadline exceeded while waiting connection")
 		}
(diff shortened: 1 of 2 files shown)


RouquinBlanc:
Second part seems to do the trick!

RouquinBlanc:
Seems fine now (running 20 minutes and no increase).

It looks like the HTTPClient keeps idle open connections for future requests and does not clean them up unless CloseIdleConnections() is called. They would eventually die out, but that depends on GC and/or remote side closing, which is not ideal.

🤞

aler9:
Thanks for this patch, i did the following changes:

* i moved `http.Client.CloseIdleConnections()` outside `WHIPClient`, since the HTTP Client is not meant as a managed component of `WHIPClient` but as an external component that may be shared with other entities, therefore `WHIPClient` should not perform actions that may interfere with other entities
* i edited `WHIPClient.Publish()` and `WHIPClient.Read()` in order to automatically call `WHIPClient.Close()`, therefore calling it manually should not be necessary anymore

I agree with you on the fact that `http.Client.CloseIdleConnections()` should be called whenever a `http.Client` is not in use anymore, i'll edit the server in order to add this in all other places where `http.Client` is in use.

This has been merged, thanks!

PR #3125, opened 2024-03-10
Create mediamtx

tschettervictor:
Simple rc service script to run mediamtx on startup on FreeBSD

No logging
Just a start and stop for now. 

Will update once I set up logging

### DIFF
--- a/rcd/mediamtx
+++ b/rcd/mediamtx
@@ -1,0 +1,22 @@
+#!/bin/sh
+
+# Mediamtx FreeBSD Service Script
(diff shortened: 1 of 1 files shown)


aler9:
Hello, sorry but we cannot merge into the main repository a script that is related to an installation procedure that is not officially supported. Usually these scripts are hosted inside dedicated repositories together with all supporting files that allow to build a package for a certain operating system, package manager or similar systems. You are free to create such repository for this specific combination.

PR #3164, opened 2024-03-25
Add option for ICE servers to be client only

### LINKED ISSUE #3145: Support asymmetric ICE
## Describe the feature
[Stunner](https://github.com/l7mp/stunner) is a kubernetes native STUN/TURN server. We can use it in symmetric ICE mode, but a nice to have would be being able to configure it in [asymmetric ICE](https://github.com/l7mp/stunner/blob/main/docs/DEPLOYMENT.md#asymmetric-ice-mode) mode when mediamtx is also running inside kubernetes. Would you be interested in accepting a pull request to enable this? I believe the changes would need to be in the signalling side so we can respond to the client with ICE configuration without using it in the server.

dbason:
Fixes #3145 

The config update will not change the default behaviour.  It does however allow the asymmetric model.  This is how LiveKit uses TURN servers for example.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1879,6 +1879,16 @@ webrtcICEServers2:
 
 where secret is the secret of the TURN server. MediaMTX will generate a set of credentials by using the secret, and credentials will be sent to clients before the WebRTC/ICE connection is established.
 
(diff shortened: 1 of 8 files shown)


dbason:
I think there might be a race condition in the failing test.  I have noticed that one other open PR is also failing on the same thing.

aler9:
merged, thanks!

PR #3170, opened 2024-03-26
Use incoming dts/pts for record and playback

bog:
- modifies the record server to not offset start of the part for dts/pts
- modifies the playback server to handle recorded files which do not start at 0 dts/pts
- modifies the beginning of recordings to not set PTSOffset/Duration to 0 of frames before the requested time, but after the last key frame. i'd like to set start_time/start_pts to the requested time on the fmp4, but not sure how to do that as of now.

Ultimately, our process uses pts/dts so we can seek to the requested values once we have the download without worrying about the extra data at the start.

### DIFF
--- a/internal/playback/fmp4.go
+++ b/internal/playback/fmp4.go
@@ -90,8 +90,9 @@ func fmp4SeekAndMuxParts(
 	maxTime time.Duration,
 	w io.Writer,
 ) (time.Duration, error) {
-	minTimeMP4 := durationGoToMp4(minTime, 90000)
-	maxTimeMP4 := durationGoToMp4(maxTime, 90000)
+	var minTimeMP4 uint64
+	var maxTimeMP4 uint64
+	var baseTime uint64
 	moofOffset := uint64(0)
 	var tfhd *mp4.Tfhd
 	var tfdt *mp4.Tfdt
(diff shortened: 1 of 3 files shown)


bog:
I don't expect this to get merged, but it better matches our workflow. We'll be testing it out.

aler9:
Hello,

> i'd like to set start_time/start_pts to the requested time on the fmp4, but not sure how to do that as of now

I tried many times to implement this, and after digging into the specification (ISO 14496-12 2015) and into the source code of FFmpeg, GStreamer and VLC i found out that this can be done by using a MP4 box called "elst" which allows to define the starting PTS of each track. I also found out that most players disable support for the elst box when they detect that the MP4 is fragmented, so it can be used with standard MP4s only. I just added support for generating standard MP4s, so this should solve your issue: https://github.com/bluenviron/mediamtx/pull/3213

> modifies the record server to not offset start of the part for dts/pts
> modifies the playback server to handle recorded files which do not start at 0 dts/pts

You shouldn't use DTS / PTS to index packets, you should instead use NTP.

The DTS / PTS shipped with incoming packets is not fit to be used inside recordings since it is protocol-dependent, in some protocols it is initialized to a random value by specification, and it gets resetted periodically (for instance, in case of RTSP, it gets resetted every 13 hours since it's a 32-bit unsigned integer, assuming a 90khz clock, and the same applies for every other protocol). Contrarily, the DTS/PTS used inside recordings is obtained by increasing a timestamp that has a very large overflow period (years).

If you use a streaming protocol that supports streaming NTP along with packets (like RTSP or HLS), and you store the NTP of the frame you're interested into somewhere, you can then use this NTP in order to retrieve the frame. Instructions on how to do this with FFMpeg are here:

https://superuser.com/questions/1470807/can-ffmpeg-obtain-the-absolute-timestamps-of-frames-captured-in-a-video-stream

Issue #3188, opened 2024-04-03
Unable to stream with RTMP and codec H265/VP9/AV1


pedroSG94:
## Which version are you using?

v1.6.0

## Which operating system are you using?

- [x] Linux amd64 standard
- [ ] Linux amd64 Docker
- [ ] Linux arm64 standard
- [ ] Linux arm64 Docker
- [ ] Linux arm7 standard
- [ ] Linux arm7 Docker
- [ ] Linux arm6 standard
- [ ] Linux arm6 Docker
- [ ] Windows amd64 standard
- [ ] Windows amd64 Docker (WSL backend)
- [ ] macOS amd64 standard
- [ ] macOS amd64 Docker
- [ ] Other (please describe)

## Describe the issue

Due to new AMF0 parser in RTMP . AMF0 Strict Array type is not supported producing this error:
```
closed: unsupported marker 0x0a
```
AMF0 documentation to implement Strict Array:
https://rtmp.veriskope.com/pdf/amf0-file-format-specification.pdf

According with RTMP enhanced in connect command you need include a property called fourCcList which is a Strict Array:
https://veovera.org/docs/enhanced/enhanced-rtmp-v1#extending-netconnection-connect-command

This was working on version 1.5.1 so I think the error is in the new AMF0 parser.
Also, I checked that if you send AMF0 types not implemented in [this method](https://github.com/bluenviron/mediamtx/blob/main/internal/protocols/rtmp/amf0/unmarshal.go#L49), the connection fail. I think that the server should at least ignore not implemented types instead of refuse connection if it is a valid type.

I did a PR to fix this error:
https://github.com/bluenviron/mediamtx/pull/3189

## Describe how to replicate the issue

1. Download the Android apk attached (this apk is a compilation of [this project](https://github.com/pedroSG94/RootEncoder) modified to work with H265 codec as default in rotation example)

[app-debug.zip](https://github.com/bluenviron/mediamtx/files/14859243/app-debug.zip)
...
2. Unzip and install the apk in an Android device
3. Open app
4. Open Rotation (include filters) example
5. Type a rtmp url in the textbox
6. Press center button to start stream
7. The connection fail and the app try reconnect.

## Did you attach the server logs?

Yes

## Did you attach a network dump?

No

aler9:
fixed by #3189

PR #3189, opened 2024-04-03
Extend AMF0 decode implementation

pedroSG94:
Hello,

I did this PR to fix the error mentioned here:
https://github.com/bluenviron/mediamtx/issues/3188
Also, I added others AMF0 types to decode extending the support:

- Add AMF0 Strict Array type decode
- Add AMF0 Long String type decode
- Add AMF0 XML Document type decode
- Add AMF0 Unsupported type decode
- Add AMF0 Undefined type decode
- Add AMF0 Date type decode

With this, the server should support all AMF0 types decode except recordedset, movieclip and typed-object.
recordedset and movieclip are reserved so you can ignore it.
typed-object is a custom object serialized so ignored it should be fine too

Note: Strict Array implementation return nil value because I'm not sure how to return an array of item. Maybe you need fix it or give me a tip for it ( I'm new with golang :( )


pedroSG94:
Fix comment error and tests added

pedroSG94:
> Hello, thank you very much for this patch - just one question, i noticed that you edited the parser in order to silently skip markerUnsupported and markerUndefined. Silently skipping things is something we've always avoided unless strictly necessary, because it makes detecting problems harder. Are you sure that this is necessary? thanks

Hello,

It is not necessary, I only did this to avoid close a connection if this types are received because the connection could be possible skipping it. Maybe a good solution could be show a warning log for debug purpose.
We can do it as you want. You know more about possible side effects, so maybe it is not a good idea to skip it anyway

aler9:
i implemented StrictArray marshaling, added fuzz tests and removed the long string / XML / date handling, since those last entities are seldom used and would require additional logic. If you encounter these entities in the future, feel free to submit another patch.

PR #3196, opened 2024-04-06
Be less harsh on B-Frames

neilyoung:
Changes the behaviour on seemingly detected B-frames (ignore instead of break)

Relevant discussion: https://github.com/bluenviron/mediamtx/discussions/2818


aler9:
Hello, the problem of this PR is that it produces corrupted streams. If you have a decoder which outputs access units with the same timestamp, open an issue and provide a sample file that allows to replicate the problem, since the server may be improved in order to merge or skip these duplicate timestamps before passing them to the output muxer.

PR #3213, opened 2024-04-08
playback: support serving streams in standard MP4 format

aler9:
## Description

This PR adds the mp4 (standard MP4) output format to the playback server, in additional to the existing fMP4 (fragmented MP4) output format.

Pros of standard MP4 with respect to fragmented MP4

* Broader compatibility
* Duration is explicit
* Frames with negative DTS are hidden by using the "edit list" feature of MP4, instead of voiding their duration, reducing warnings and errors from some players

Cons

* Generation is slower

## Usage

Add `format=mp4` to paths in playback server:

```
http://localhost:9996/get?path=stream&start=2024-04-07T12%3A57%3A33.808301%2B02%3A00&duration=1147&format=mp4
```

## Nightly binaries

https://github.com/bluenviron/mediamtx/actions/runs/8621626444

## TODO

* [X] sort samples of different tracks by DTS before writing them
* [ ] Tests
* [x] Documentation

### DIFF
--- a/README.md
+++ b/README.md
@@ -1348,7 +1348,7 @@ Where [mypath] is the name of a path. The server will return a list of timespans
 The server provides an endpoint for downloading recordings:
 
 ```
-http://localhost:9996/get?path=[mypath]&start=[start_date]&duration=[duration]
+http://localhost:9996/get?path=[mypath]&start=[start_date]&duration=[duration]&format=[format]
 ```
 
 Where:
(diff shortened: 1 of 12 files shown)


stardast:
Hi, a question about the cons:
What do you mean by "generation is slower"? Is a file created that concatenates all the segments? On disk or in RAM? or is it necessary to read all the segments to get the total size without creating a temporary file? Is the slowness proportional to the length of the requested video?

Thanks,
Paolo

aler9:
> What do you mean by "generation is slower"? Is a file created that concatenates all the segments? On disk or in RAM? or is it necessary to read all the segments to get the total size without creating a temporary file? Is the slowness proportional to the length of the requested video?

The slowness is caused by the fact that in standard MP4s there is a single list that contains all samples. Therefore, all segments must be scanned and their samples have to be parsed before the playback file can be served.

Yes, the slowness is proportional to the length of the requested video.

In fragmented MP4 on the other hand there are multiple, small lists of samples, and these are concatenated them together along with the sample content. Therefore, the playback file can be generated as soon as the first segment is scanned.

PR #3253, opened 2024-04-16
fix support for JWT authentication in API, metrics, playback, pprof

rfscheidt:
I've made some modifications related to JWT authentication that I would like to bring to your attention. It appears that token validation was not being performed in the methods listed below. Additionally, I attempted to integrate it with HLS, but encountered some issues. It might be beneficial to review these changes to ensure everything is functioning as expected.

Modified Files:

internal/api/api.go
internal/metrics/metrics.go
internal/playback/server.go
internal/pprof/pprof.go

Your feedback and review on these changes would be greatly appreciated. Please let me know if there are any concerns or if further adjustments are needed.

Thank you for your attention to this matter.


rfscheidt:
@aler9 do you have time do review this PR?

aler9:

PR #3255, opened 2024-04-17
fix when runOnDemand for HLS, variable MTX_QUERY is empty.

xjr20042:
# The following environment variables are available:
...
# * MTX_QUERY: query parameters (passed by first reader)
runOnDemand:
...
The comments in the configuration file say that MTX_QUERY is available, but in fact, it is empty when the HLS stream is requested.
I passed the query from the hls/http_server to the muxer.


aler9:
merged, thanks!

PR #3299, opened 2024-04-27
fix VP9 decoding on iOS (#3060)

aler9:


aler9:
this will be applied automatically during the next gortsplib upgrade.

PR #3302, opened 2024-04-29
fix rtmp server miss the opportunity to read metadata when clients send set chunk size message after publishing stream.

xjr20042:
some rtmp client will send set chunk size after publishing stream, which causes the metadata to fail to be read.
I add code to skip set chunk size message in readTracks() in rtmp/reader.go.


aler9:
merged, thanks!

Issue #3361, opened 2024-05-16
Replace path API endpoint shouldn't error if the path doesn't exist currently


bouk:
## Describe the feature

Currently the replace path endpoint return 404 when setting an endpoint that doesn't exist currently, which doesn't really serve any purpose since the path is replaced anyways. This is annoying for clients, since they have to loop to try creating the path instead which can then also fail, because it might've been created in the meantime.

Making it accept replace for paths that don't exist yet will make it idempotent

PR #3362, opened 2024-05-16
api: make replace paths idempotent

bouk:
Don't return a 404 error if the path doesn't exist yet

Fixes https://github.com/bluenviron/mediamtx/issues/3361


aler9:
merged, thanks!

PR #3373, opened 2024-05-19
migrate to latest version

udhay24:


aler9:
it's always rewarding to know that people are using MediaMTX as the core component of other platforms :)

PR #3388, opened 2024-05-24
fix error in timestamp of Opus packets coming from SRT / HLS / MPEG-TS

aler9:


bouk:
Incredible, I was trying to get SRT -> WebRTC opus audio working on you already fixed it 🤩

Issue #3404, opened 2024-05-28
Make webrtcHandshakeTimeout and webrtcTrackGatherTimeout configurable


jwalton:
## Describe the feature

I'm running into an issue where my WebRTC client is taking a little too long to complete the timeout.  We end up hitting the `webrtcHandshakeTimeout`, which kills the connection, then my client tries again, but fails again, and we end up in this loop of doom never playing video, but it could be solved if we could bump up the timeout a little.

So I'd like to raise a PR to make this configurable, but I'm wondering if you would prefer to see these added to the `Path` struct (so they could be configured per-path) or to the top level `Conf` struct?  I'm inclined towards the `Conf` struct, as it seems unlikely anyone would ever want to configure these path-by-path.

aler9:
fixed by #3406

PR #3406, opened 2024-05-29
webrtc: allow configuring timeouts (#3404)

jwalton:
Allow configuration of WebRTC timeouts.


internal/core/core.go
aler9, line 1: in line 853, add a comparison between current and previos WebRTCHandshakeTimeout p.conf.WebRTCTrackGatherTimeout in order to detect changes after a API request or hot reloading

aler9 (changes requested):
thanks for this patch, take a look at the comment in the review

internal/core/core.go
jwalton: Fixed.

aler9 (approved):

aler9:
merged, thanks!

Issue #3430, opened 2024-06-07
JWT authentication token too long for SRT publish 


dm-dma:
## Which version are you using?

v1.8.2

## Which operating system are you using?

- [ ] Linux amd64 standard
- [ ] Linux amd64 Docker
- [ ] Linux arm64 standard
- [ ] Linux arm64 Docker
- [ ] Linux arm7 standard
- [ ] Linux arm7 Docker
- [ ] Linux arm6 standard
- [ ] Linux arm6 Docker
- [x] Windows amd64 standard
- [ ] Windows amd64 Docker (WSL backend)
- [ ] macOS amd64 standard
- [ ] macOS amd64 Docker
- [ ] Other (please describe)

## Describe the issue

In order to send the JWT token using SRT when JWT authentication is enabled, the JWT token is a part of the stream ID:
srt://x.x.x.x:8890?streamid=publish:mystream:jwt=[token]

The issue with this implementation is that the streamid may only contain 512 characters. We created a token with almost the bare minimum but ended up with 685 characters:
https://github.com/Haivision/srt/blob/master/docs/features/access-control.md

So technically this is not a bug and this is SRT related. But this leaves us with an implementation that is unusable, or is there another way to pass JWT tokens?

## Describe how to replicate the issue

1. start the server
2. publish with ...
```
./srt-live-transmit udp://:5004 "srt://192.168.5.99:8890?streamid=publish:mystream:jwt=[token]" -v -s 100 -ll debug
Media path: 'udp://:5004' --> 'srt://192.168.5.99:8890?streamid=publish:mystream:jwt=[token]'
SRT parameters specified:

        streamid = 'publish:mystream:jwt=[token]'
Opening SRT target caller on 192.168.5.99:8890
16:52:24.463124/srt-live-transm D:SRT.sm: generateSocketID: : @310547703
WARNING: failed to set options: streamid,

ERROR #5003.0: Operation not supported: Bad parameters
SrtCommon: DESTROYING CONNECTION, closing sockets (rt%310547703 ls%-1)...
SrtCommon: ... done.
ERROR: error: ConfigurePre: Operation not supported: Bad parameters
```

## Did you attach the server logs?

no

## Did you attach a network dump?

no

jquiroga-arg:
Hi, I implemented a workaround to address JWT token length limitations present in several clients, such as FFmpeg. This approach achieves around a 30% reduction in token size. What I do is compress the token using Gzip, and then decompress it from MediaMTX.

[...]

PR #3431, opened 2024-06-07
Exclude actions when using JWT authentication

dm-dma:
In order to overcome the issue in #3430 and probably for other use cases as well I added an authJWTExclude parameter that functions the same as authHTTPExclude. So actions can be excluded from the JWT authentication


aler9:
merged, thanks and sorry for the time it took to review.

PR #3464, opened 2024-06-13
Save a hash of the hls.js release

bouk:
This ensures the downloaded hls.js matches exactly and removes a dependency on cdn.jsdelivr.net

It also makes it easier for the nixpkgs version of mediamtx to get the hls.min.js updates, because [right now this needs to be done manually](https://github.com/NixOS/nixpkgs/blob/3f84a279f1a6290ce154c5531378acc827836fbb/pkgs/servers/mediamtx/default.nix#L9-L12)


aler9:
This has been merged, thanks.

Regarding the NixOS problem, in my opinion it has been fixed in the wrong way, since the proper solution consists into adding `go generate` to the package building process. Command `go generate` was introduced in order to run any pre-compilation step by using a single, standard, command. If pre-compilation steps are instead embed into the package and run manually, there's no guarantee that the software will compile correctly in the future.

PR #3489, opened 2024-06-19
Add start filter for list endpoint

tomik148:
Add non required "start" filter to list endpoint, works same as in get endpoint but is optional, if not provided endpoint works same as before if is provided only returns segments created after specified datetime.

### DIFF
--- a/internal/playback/on_list.go
+++ b/internal/playback/on_list.go
@@ -7,6 +7,7 @@ import (
 	"io"
 	"net/http"
 	"os"
+	"slices"
 	"time"
 
 	"github.com/bluenviron/mediacommon/pkg/formats/fmp4"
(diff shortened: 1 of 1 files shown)


aler9:
Hello, while the intended feature is quite useful, the implementation is not correct since it works with segments, while the aim of the playback server is abstracting segments and working on top of them. Here's an example: if we have 3 segments, each 1 hour long:

```
09_00_00.mp4
10_00_00.mp4
11_00_00.mp4
```

calling `/list` results in a single entry since segments are concatenated:

```
[
  {
    "start": "09_00_00",
    "duration": "10800.0"
  }
]
```

if we call `/list?start=09:30:00` with your implementation, the result is

```
[
  {
    "start": "10:00:00",
    "duration": "7200.0"
  }
]
```

This is not correct since there's a recording between 9:30 and 10:00, that is part of the segment that is wrongly filtered out. Correct result should be:

```
[
  {
    "start": "09:30:00",
    "duration": "9000.0"
  }
]
```

Therefore, try reimplementing the feature, for instance, by hacking `FindSegments()` and `computeDurationAndConcatenate()` in order to work with a start date.

aler9:
replaced by #4085

Issue #3455, opened 2024-06-11
runOnDemandCloseAfter blocks the path after teardown


mrlt8:
## Which version are you using?

v1.8.2

## Which operating system are you using?

- [ ] Linux amd64 standard
- [ ] Linux amd64 Docker
- [ ] Linux arm64 standard
- [x] Linux arm64 Docker
- [ ] Linux arm7 standard
- [ ] Linux arm7 Docker
- [ ] Linux arm6 standard
- [ ] Linux arm6 Docker
- [ ] Windows amd64 standard
- [ ] Windows amd64 Docker (WSL backend)
- [ ] macOS amd64 standard
- [ ] macOS amd64 Docker
- [ ] Other (please describe)

## Describe the issue

Using `runOnDemand` with `runOnDemandCloseAfter` seems to create an issue where the path will block and become unresponsive until the `runOnDemandCloseAfter` times out even if the path is destroyed.

Can we get `runOnDemandCloseAfter` to reset and run immediately if the path is destroyed or a way to reset it? 

Or is there a way to clear out and reset the path with `runOnNotReady`?

## Describe how to replicate the issue

1. Set `runOnDemandCloseAfter: 60s`
2. start the server
3. read the path
4. stop the path source
5. attempt to read the path again before the timeout
6. attempt to read the path again after the timeout

## Did you attach the server logs?

```
<!-- STOP SOURCE -->

07:07:45 DEB [RTSP] [conn 127.0.0.1:37054] [c->s] TEARDOWN rtsp://0.0.0.0:8554/front-door RTSP/1.0
CSeq: 6
Session: 6e3581c500424bef8d4695af4e4ac9a8
User-Agent: Lavf60.3.100


07:07:45 INF [path front-door] runOnReady command stopped
07:07:45 INF [path front-door] runOnNotReady command launched
07:07:45 INF [RTSP] [session de560115] destroyed: torn down by 127.0.0.1:37054
07:07:45 DEB [RTSP] [conn 127.0.0.1:37054] [s->c] RTSP/1.0 200 OK
CSeq: 6
Server: gortsplib


07:07:45 INF [RTSP] [conn 127.0.0.1:37054] closed: EOF


<!-- ATTEMPT TO OPEN PATH FAILS -->
07:08:15 INF [RTSP] [conn 192.168.65.1:53748] opened
[...]

PR #3507, opened 2024-06-25
onDemandPublisherStop when publisher removed  #3455

mrlt8:
This should call `onDemandPublisherStop` when a publisher is removed instead of waiting for `onDemandPublisherCloseTimer` to timeout which would cause the path to block if `runOnDemandCloseAfter` is set (#3455).


aler9:
Thanks for the patch, but read here: https://github.com/bluenviron/mediamtx/issues/3455#issuecomment-2267098247

PR #3560, opened 2024-07-18
Configure the JWT Claim Key for holding MediaMTX Permissions

izanbard:
### Why is this change needed?
I offer this change as when using Azure B2C the policy for naming custom JWT claims is to prefix with `extension_<<app_id>>_`.  With MediaMTX having a fixed JWT of `mediamtx_permissions` it meant that I could not use Azure B2C as my IdP.

This will be the case for many people who need to secure video streams using JWTs. 

### What has been done?
The main changes are in `manager.go:268-283` and `conf.go:568-573`.  

- `manager.go`: I have intercepted the automagic un-marshalling of the JWT in to `customClaims` object and done it manually instead.  This means that the key used to search the claim for the `mediamtx_permissions` can be set at runtime rather than at compile time.
- `conf.go`: in addition to setting the default for the new config key (`authJWTClaimKey`) to be `mediamtx_permissions` for backward compatibility, I have also added some validation to ensure the claim key is not blank and that it only has valid JWT chars in it `[a-zA-Z0-9-_]`.

I have updated unit tests and made simple changes to the `README.md` and default `mediamtx.yml` files.


izanbard:
@aler9 The test above that has failed is nowhere near the code I have written; passes on my system and is similar to a fail you had yesterday that went away on retry.  Please rerun the check.

izanbard:
@aler9 thank you

internal/auth/manager.go
```diff
@@ -269,13 +265,26 @@ func (m *Manager) authenticateJWT(req *Request) error {
...
 
-	var cc customClaims
-	_, err = jwt.ParseWithClaims(v["jwt"][0], &cc, keyfunc)
+	token, err := jwt.Parse(v["jwt"][0], keyfunc)
 	if err != nil {
 		return err
 	}
 
-	if !matchesPermission(cc.MediaMTXPermissions, req) {
+	claims := token.Claims.(jwt.MapClaims)[m.JWTClaimKey].([]interface{})
```
aler9, line 273: this is prone to panics since `token.Claims.(jwt.MapClaims)[m.JWTClaimKey]` can be `nil`

internal/auth/manager.go
```diff
@@ -269,13 +265,26 @@ func (m *Manager) authenticateJWT(req *Request) error {
...
+	MediaMTXPermissions := make([]conf.AuthInternalUserPermission, 0, len(claims))
+
+	for _, permission := range claims {
+		var MediaMTXPermission conf.AuthInternalUserPermission
+		MediaMTXPermission.Action = conf.AuthAction(permission.(map[string]interface{})["action"].(string))
```
aler9, line 278: this is also a possible cause of panics since permission might not be a `(map[string]interface{})`

aler9 (changes requested):
Hello, this feature is useful but you need to perform a couple of changes in order to get the patch merged:
* the permission parsing logic is vulnerable to multiple crashes. In my opinion you should rewrite it entirely by using something like

```go
type customClaims map[string]json.RawMessage
var cc customClaims
_, err = jwt.ParseWithClaims(v["jwt"][0], &cc, keyfunc)
...

rawClaim, ok := cc[m.JWTClaimKey]
if !ok {
   return fmt.Errorf(...)
}

var claim []conf.AuthInternalUserPermission
err := json.Unmarshal(rawClaim, &claim)
if err != nil {
   return err
}

// finally you can use the claim
```
* you need to add the new key to `apidocs/openapi.yaml` too

aler9:
i found a way to implement this feature that is slightly more efficient, since it avoids calling `json.Marshal` and `json.Unmarshal` in series. This has been moved here: #3692

PR #3598, opened 2024-08-02
Dynamically refresh tls certs for all servers

### LINKED ISSUE #3570: Automatically reload when SSL certs change
## Describe the feature
We have a system that automatically rotates SSL certs that are used in our RTMPS endpoint, however when the certs get updated they are not updated in mediamtx without restarting the service.  It would be a big QoL feature for the service to automatically reload these when they change.

dbason:
Fixes #3570


aler9:
merged, thank you very much!

PR #3622, opened 2024-08-05
remove whitespace from hashBuf

San9H0:
In Windows, `go generate ./...` was causing a `ERR: hash mismatch` error.

To resolve this issue, I added a line to trim any whitespace from the `hashBuf`

errors no longer occur


aler9:
merged, thanks!

PR #3663, opened 2024-08-16
perf: chunk and process playback segments concurrently

boenshao:
The `/list` API can be a bit slow when there are lots of recordings that need to be processed.

The concatenation procedure can be split up and executed concurrently (and even parallelly).

I hard-coded the splits into 4, as most modern platforms have at least 4 cores. 

But it's possible to introduce a more sophisticated way to determine the preferred number, either by the number of files or the total size of the files. We can gather the statistics in `os.Walk`.

Yet I believe the overhead of goroutine is negligible in most cases, so a static strategy is probably alright.

With this patch, on my AlderLake laptop, the response time is about 3x faster when parsing 2X GBs of recordings.

### DIFF
--- a/internal/playback/on_list.go
+++ b/internal/playback/on_list.go
@@ -8,6 +8,7 @@ import (
 	"net/http"
 	"net/url"
 	"os"
+	"sort"
 	"strconv"
 	"time"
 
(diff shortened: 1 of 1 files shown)


aler9:
This PR has an additional issue: it prevents more than 4 segments from being grouped together. The `/list` endpoint is supposed to provide timespans as long as possible, in order for them to be represented in a UI, not the segment list, which can be quite large.

This is replaced by #4102 which parses segments in parallel but performs concatenation in series.

PR #3688, opened 2024-08-24
KLV metadata handling

rmcnew:
Changes to support KLV metadata handling in mediamtx.

See related KLV metadata pull requests in mediacommon, gortsplib, and mediamtx.

All these changes released to public domain for adding to parent project under parent project's license(s).

### DIFF
--- a/internal/formatprocessor/klv.go
+++ b/internal/formatprocessor/klv.go
@@ -1,0 +1,134 @@
+package formatprocessor
+
+import (
+	"fmt"
+	"time"
+
+	"github.com/bluenviron/gortsplib/v4/pkg/format"
+	"github.com/bluenviron/gortsplib/v4/pkg/format/rtpsimpleaudio"
+	"github.com/bluenviron/gortsplib/v4/pkg/rtptime"
+	"github.com/pion/rtp"
+
+	"github.com/bluenviron/mediamtx/internal/unit"
+)
+
+type formatProcessorKLV struct {
+	udpMaxPayloadSize int
+	format            *format.KLV
+	timeEncoder       *rtptime.Encoder
+	encoder           *rtpsimpleaudio.Encoder
+	decoder           *rtpsimpleaudio.Decoder
+}
+
+func newKLV(
+	udpMaxPayloadSize int,
+	forma *format.KLV,
+	generateRTPPackets bool,
+) (*formatProcessorKLV, error) {
+	t := &formatProcessorKLV{
+		udpMaxPayloadSize: udpMaxPayloadSize,
+		format:            forma,
+	}
+
+	if generateRTPPackets {
+		err := t.createEncoder()
+		if err != nil {
+			return nil, err
+		}
+
+		t.timeEncoder = &rtptime.Encoder{
+			ClockRate: forma.ClockRate(),
+		}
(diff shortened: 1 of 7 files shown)


rmcnew:
This pull request depends on changes in the mediacommon (https://github.com/bluenviron/mediacommon/pull/145) and gortsplib  (https://github.com/bluenviron/gortsplib/pull/606).

rmcnew:
Thank you for looking at my pull requests.

aler9:
i've reviewed the whole thing, implementation is almost complete, here are the missing steps that i spotted:

* in gortsplib, when decoding/encoding KLV from/to RTP packets, it's not enough to just take the payload of every RTP packet and treat is as a single KLV item (as rtpsimpleaudio does) since a KLV item might be splitted into multiple packets. The right decoding procedure consists into obtaining the KLV item size by looking at the value length field (3rd byte) and use that value to concatenate every incoming RTP payload until the unit is complete. Reference implementation:
https://gitlab.freedesktop.org/gstreamer/gst-plugins-good/-/blob/master/gst/rtp/gstrtpklvdepay.c

   furthermore, the specification says to add a Marker bit to to the last packet of the last KLV item belonging to a certain timestamp, so i fear that we will be forced to group together all KLV items with the same timestamp and route them together all along the server and all libraries. The specification says that a group of KLV items is called a KLV unit.

   Therefore: there must be a dedicated RTP/KLV decoder and encoder that is able to read/write KLV units from/to multiple RTP packets. 

* in mediacommon, i understand that KLV in MPEG-TS might not have a PTS, but this prevents KLV from being muxed into RTP. The right approach consists into improving the MPEG-TS reader to store the last valid PTS and using that instead when PTS is not present.

* in mediacommon, you added code to decode metadata but you did not use it (pkg/formats/mpegts/KLVA_access_unit.go) - in this way it's not possible to decode KLV properly (when stream type is metadata). You have to use it when StreamType is metadata.

rmcnew:
Thank you for your review and feedback.

I will work on the issues and submit a new pull request later.

Sanderi44:
@rmcnew I'm still learning mediamtx and have a question about this. If I have an RTSP + MPEGTS stream with KLV, is it possible, with the current config file, to strip the KLV data and rebroadcast to a new RTSP url? It would be great to be able to view the HLS stream, as well. Do you know how to do that?

aler9:
@Sanderi44 the feature you are describing is already implemented and is the default behavior.

KLV tracks are currently ignored.
MPEG-TS to RTSP conversion is automatic, as all conversions.
And here you have it.

By the way, if @rmcnew is able to provide the small parts that were missing from his implementation, I'm willing to merge everything.

rmcnew:
As soon as the needed work is completed and tested, I will open another pull request.  Apologies, I have other urgent tasks that I must complete first.

If someone else is able to add the missing parts and create a pull request, that would be good.

rmcnew:
> @rmcnew Hi, Are there any improvements regarding encoding/decoding "RTSP + MPEGTS (KLV)" that can be committed ? Thanks.

I do not have anything ready right now.  Anyone can take the changes from my incomplete pull request and use them to create a complete solution as Alessandro described above.

Issue #3155, opened 2024-03-19
HLS streaming playback speed is 1.5x for mobile devices with no option to change


tschettervictor:
## Which version are you using?

v1.6.0

## Which operating system are you using?

- [ ] Linux amd64 standard
- [x] Linux amd64 Docker
- [ ] Linux arm64 standard
- [ ] Linux arm64 Docker
- [ ] Linux arm7 standard
- [ ] Linux arm7 Docker
- [ ] Linux arm6 standard
- [ ] Linux arm6 Docker
- [ ] Windows amd64 standard
- [ ] Windows amd64 Docker (WSL backend)
- [ ] macOS amd64 standard
- [ ] macOS amd64 Docker
- [x] Other (please describe) FreeBSD 13.2

## Describe the issue

When viewing a stream via HLS from mobile devices (ios ipados safari,brave,chrome etc) the playback speed will default to 1.5x and cannot be changed for it seems the first 10 seconds of most recent playback. If you skip back 10 seconds, or pause until there is a 10 second gap, the speed will go back to 1x also without the ability to change.
The HLS stream works fine for desktop using brave browser. Other browsers are not tested

This is with all three variants of HLS, and storing in RAM or directory.

## Describe how to replicate the issue

1. start the server
2. publish with rtmp any stream with user that has publish permissions
3. read with hls from mobile devices (ios, android not tested) ios version 17.3.1 and 17.4

## Did you attach the server logs?

no.
There are no related logs. Logs all show properly...stream started and is muxing to HLS

## Did you attach a network dump?

no

tschettervictor:
Just tested with the latest docker image. Same issue.

tschettervictor:
@aler9 Do you think this is an issue with the HLS player, or the devices themselves. I read that you can set “playbackrate” for HLS but haven’t been able to solve the issue yet.

tschettervictor:
Any thoughts on this. Its fairly important to get this functioning properly, otherwise I’ll look elsewhere for a streaming solution.

jwetzell:
This is reproducible still in 1.8.3. I can reproduce it in some very specific scenarios (navigate away and back quickly or start/stopping the player quickly) on both mobile and desktop. This has to do with this line [here](https://github.com/bluenviron/mediamtx/blob/afec0cd86cd07375284cdac9c330743e2d58f667/internal/servers/hls/index.html#L72) as I can tweak the playback speed that the HLS player bumps to by changing that line.

[...]

PR #3731, opened 2024-09-04
HLS: iOS clients treat maxLiveSyncPlaybackRate as the default playback rate

deepseth:
As highlighted in #3155, iOS clients are treating maxLiveSyncPlaybackRate as the default playback speed. Changing this to 1.0 (from 1.5) results in iOS clients streaming at 1.0 speed by default instead.

### DIFF
--- a/internal/servers/hls/index.html
+++ b/internal/servers/hls/index.html
@@ -69,7 +69,7 @@ const loadStream = () => {
 	// but don't support fMP4s.
 	if (Hls.isSupported()) {
 		const hls = new Hls({
-			maxLiveSyncPlaybackRate: 1.5,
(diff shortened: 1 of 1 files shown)


aler9:
@deepseth thanks for the explanation, i just found out why i was not able to replicate the issue: it arises when using iOS with a hls.js-capable device. Usually, iOS devices are not able to run hls.js since they miss the HTML5 Media Source extension. This is related to the iOS + iPad combination.

The only problem i see is that `maxLiveSyncPlaybackRate` must be set to 1.5 when running with any other device, in order to compensate for delays during transmission, therefore there must be a way to differentiate between the two cases.

aler9:
replaced by #3734

PR #3734, opened 2024-09-04
hls: fix playback speed on iOS (#3155) (#3731)

### LINKED ISSUE #3155: HLS streaming playback speed is 1.5x for mobile devices with no option to change

aler9:
Fixes #3155 
Replaces #3731

### DIFF
--- a/internal/servers/hls/index.html
+++ b/internal/servers/hls/index.html
@@ -63,11 +63,18 @@ const setMessage = (str) => {
 	message.innerText = str;
 };
 
+const isIOS = () => (
+	/iPad|iPhone|iPod/.test(navigator.platform)
+	|| (navigator.platform === 'MacIntel' && navigator.maxTouchPoints > 1)
+);
+
 const loadStream = () => {
-	// always prefer hls.js over native HLS.
-	// this is because some Android versions support native HLS
+	// Prefer hls.js over native HLS.
+	// This is because some Android versions support native HLS
 	// but don't support fMP4s.
-	if (Hls.isSupported()) {
+	// Skip iPad iOS >= 13 and iPhone iOS >= 17,
+	// which support hls.js but don't support well maxLiveSyncPlaybackRate.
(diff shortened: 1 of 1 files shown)


deepseth:
To note that the device I was testing on iOS with was an iPhone 13 mini running iOS 17.6.1, not an iPad.

aler9:
@deepseth thanks for the feedback, i looked around and it seems like iPhone started supporting Media Source Extensions (and therefore hls.js) since iOS 17, but maxLiveSyncPlaybackRate is still causing troubles, therefore it's better to use the native HLS implementation.

deepseth:
Super, thank you for looking at this so quickly!

Issue #3510, opened 2024-06-26
Playback across segments


MengLi619:
## Describe the feature

Description

There is a requirement for us to playback near a time point (+/- 30s), but sometimes the stream will break, and multiple segments are created, the mediamtx only playback the first segment, which may not contain the desired time point. I don't know if there is any possibility to play all segments in the time range, regardless of any break in the middle.

Thanks a lot and appreciate your excellent work.

Mediamtx 1.5.1

moenodedev:
Discussion #3243

aler9:
Hello, we provide the `/list` endpoint that allows to list available time spans and allows users to build their own player around them and to manage gaps between segments in a personalized way.

If you want to skip gaps, then as mentioned by @moenodedev there's #3243

If you want to fill gaps, this is unimplemented but tracked in #2214 

If you want to fill gaps without waiting for the new feature, build a custom player that shows a blank screen during gaps.

Issue #2214, opened 2023-08-16
Allow a publisher to reconnect without disconnecting readers


aler9:
## Describe the feature

Blank screen or fallback stream should be routed to readers. Blank screen or fallback streams are checked for codec before being used in place of the original stream.

In case the publisher reconnects, the new stream should be checked:

* if tracks are equal to the one of the original stream, the new stream is routed to readers
* otherwise, readers are resetted

rnuck2:
This feature could be very useful to me. Is its implementation in progress?
At the moment, with the latest release (builded from main branch code), i get alwaysAvailable field not existing.
Thank you.

mrasvialpha:
It would also be good if there is an option to wait until a timeout is reached to check if the publisher reconnects and continue sending frame. Currently all the readers are disconnected immediately.

efraimjer:
any news on this feature?

xaionaro:
Also curious if there are any plans to make it happen anytime soon :)

JoLoZ:
Are there any updates regarding this feature?

winkmichael:
If I understand this correct, this would give the user the ability to have a stream and insert a "No Video.' or "No Input Available" type message into the stream itself?

epunter:
This would be an incredibly good feature, currently the only thing I need to get my gold standard home RTSP server

m1k1o:
I would like to implement this feature, because I am looking for a way how to acheive exactly this behavior for RTMP ingest. While my research I found out that [liquidsoap has already something experimental like this issue](https://www.liquidsoap.info/doc-2.1.2/ffmpeg_cookbook.html#live-switch-between-encoded-content) and there is a way how to acheive it with [gstreamer in rust](https://github.com/centricular/rtmp-slate-fallback).

@aler9 is there any idea / hints from your side how to implement this freatue? Some information that would be beneficial for me to know before I fork this repository and start tinkering with possible solutions? Thanks.

TagorePde:
mistserver , have failover streaming feature.

alex-rsk:
Hello, is this feature already implemented?

NextWork123:
hello,  when that feature gonna implement?

devnoname120:
Sorry for going a bit of a tangent but does anyone here know of a similar tool that has that feature? Either self-hosted or SaaS.

fragtion:
[...]

PR #3816, opened 2024-09-28
feat: playback can skip gap between segments

MengLi619:
Related to #3510, It will be very useful to concat record segments by Mediamtx. In our project, the network is not stable and will generate a lot of small segments. Also, we need to playback the video in real time. It's very difficult for us to concat the segments out of mediamtx, this is the PR to concat the segments when playback by the mediamtx.

I don't know if this pull request can be merged, please help to review it, thanks a lot!

### DIFF
--- a/internal/conf/conf.go
+++ b/internal/conf/conf.go
@@ -207,13 +207,14 @@ type Conf struct {
 	PPROFTrustedProxies IPNetworks `json:"pprofTrustedProxies"`
 
 	// Playback
-	Playback               bool       `json:"playback"`
-	PlaybackAddress        string     `json:"playbackAddress"`
-	PlaybackEncryption     bool       `json:"playbackEncryption"`
-	PlaybackServerKey      string     `json:"playbackServerKey"`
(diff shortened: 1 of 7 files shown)


dillanspencer:
I also need this feature. Specifically, I want the ability to download video recordings from any time within the last 24 hours, even if the requested duration spans multiple recording segments. For example, if a 15-minute event occurred, I need to be able to retrieve an exact 15-minute video, regardless of whether it falls within a single recording segment or across multiple segments.

aler9:
@dillanspencer the feature you described is already implemented.
This PR is related to gaps caused by multiple starts and stops of streams.

dillanspencer:
@aler9 Ok I might have an issue then. It seems that when I download large video clips it causes an error that creates a new segment. Then when I re-request video it only gives me video from the previous segment. Should I make an issue for this?

aler9:
This will not be merged. If you need smooth play back, you can either:
* build a custom JavaScript interface that handles gaps
* wait for #2214 
* improve network throughput

PR #3852, opened 2024-10-08
versiongetter: use a permissive AlternatesFS with git-go

database64128:
versiongetter does not work in Arch Linux's package build environment, because `repo.Log(&git.LogOptions{From: head.Hash()})` somehow returns `plumbing.ErrObjectNotFound`.

It might take some time before I can find out the root cause and submit a fix, so here's the preliminary change that wraps errors with additional context.

**Update: See the added comment in code.**

Fixes #3409.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -16,6 +16,7 @@ require (
 	github.com/fsnotify/fsnotify v1.7.0
 	github.com/gin-contrib/pprof v1.5.0
 	github.com/gin-gonic/gin v1.10.0
+	github.com/go-git/go-billy/v5 v5.5.0
 	github.com/go-git/go-git/v5 v5.12.0
 	github.com/golang-jwt/jwt/v5 v5.2.1
 	github.com/google/uuid v1.6.0
@@ -54,7 +55,6 @@ require (
 	github.com/gabriel-vasile/mimetype v1.4.3 // indirect
 	github.com/gin-contrib/sse v0.1.0 // indirect
 	github.com/go-git/gcfg v1.5.1-0.20230307220236-3a3c6141e376 // indirect
-	github.com/go-git/go-billy/v5 v5.5.0 // indirect
 	github.com/go-playground/locales v0.14.1 // indirect
 	github.com/go-playground/universal-translator v0.18.1 // indirect
 	github.com/go-playground/validator/v10 v10.20.0 // indirect
--- a/internal/core/versiongetter/main.go
+++ b/internal/core/versiongetter/main.go
@@ -2,6 +2,7 @@
 package main
 
 import (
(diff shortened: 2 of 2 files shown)


aler9:
Hello, the error message you've reported is usually triggered when repositories are pulled with `--depth=1`:

```
git clone --depth=1 https://github.com/bluenviron/mediamtx
```

When repositories are cloned in this way, the tag list is not available:

```
git tag
```

(this returns an empty list)

and the server is unable to get its tag.

In my opinion there are two ways to solve the issue:

1. we can find an alternative way to get the server tag, that is compatible with `--depth=1`
2. you have to alter the Arch Linux package in order to download the full repository.

I like the first approach better - try looking around.

database64128:
> you have to alter the Arch Linux package in order to download the full repository.

AFAIK this is not possible. The `makepkg` program first clones the remote repository into a local mirror repository, and then clones the local mirror repository into a separate package `src` repository, and switches to a new `makepkg` branch, where the actual build script runs. There's no way to change how the `src` repository is set up.

Mirror repo config (`mediamtx/config`):

```gitconfig
[core]
    repositoryformatversion = 0
    filemode = true
    bare = true
[remote "origin"]
    url = https://github.com/bluenviron/mediamtx.git
    fetch = +refs/*:refs/*
    mirror = true
```

`src` repo config (`src/mediamtx/.git/config`):

```gitconfig
[core]
    repositoryformatversion = 0
    filemode = true
    bare = false
    logallrefupdates = true
[remote "origin"]
    url = /home/database64128/repos/aur/mediamtx/mediamtx
    fetch = +refs/heads/*:refs/remotes/origin/*
[branch "main"]
    remote = origin
    merge = refs/heads/main
```

Inside the `src` repo, `git log` correctly shows the latest commit/tag:

```
commit 32d3fc55ccc631ad125462063b7bf387595209fe (HEAD -> makepkg, tag: v1.9.2)
```

database64128:
> When repositories are cloned in this way, the tag list is not available:

The `src` repo is not cloned with `--depth=1`. The full tag list and commit history are available via `git tag` and `git log`.

`repo.Tags()` is actually able to iterate over all tags. `repo.Log(&git.LogOptions{From: head.Hash()})` does not work and `repo.Log(&git.LogOptions{All: true})` returns an iterator that immediately gives `EOF`.

database64128:
Likely the same issue as mentioned in https://github.com/go-git/go-git/discussions/1125 and https://github.com/go-git/go-git/issues/1002. I'll update this PR once I get things working.

database64128:
Hi, I have pushed the fix to the branch and updated the PR title. It's now ready for review.

PR #3899, opened 2024-10-22
Restart ICE only on failed connection states

andresuribe87:
According to https://www.w3.org/TR/webrtc/#dictionary-rtcofferoptions-members ,

> Note
> Performing an ICE restart is recommended when [iceConnectionState](https://www.w3.org/TR/webrtc/#dom-peerconnection-ice-connection-state) transitions to "[failed](https://www.w3.org/TR/webrtc/#dom-rtciceconnectionstate-failed)". An application may additionally choose to listen for the [iceConnectionState](https://www.w3.org/TR/webrtc/#dom-peerconnection-ice-connection-state) transition to "[disconnected](https://www.w3.org/TR/webrtc/#dom-rtciceconnectionstate-disconnected)" and then use other sources of information (such as using [getStats](https://www.w3.org/TR/webrtc/#widl-RTCPeerConnection-getStats-Promise-RTCStatsReport--MediaStreamTrack-selector) to measure if the number of bytes sent or received over the next couple of seconds increases) to determine whether an ICE restart is advisable.

This PR changes the default webrtc implementation to follow such guidance. We've seen instances where choppy networks trigger a lot of renegotiations that aren't necessary.

### DIFF
--- a/internal/protocols/webrtc/peer_connection.go
+++ b/internal/protocols/webrtc/peer_connection.go
@@ -86,7 +86,7 @@ type PeerConnection struct {
 	stateChangeMutex  sync.Mutex
 	newLocalCandidate chan *webrtc.ICECandidateInit
 	connected         chan struct{}
-	disconnected      chan struct{}
+	failed            chan struct{}
 	done              chan struct{}
 	gatheringDone     chan struct{}
 	incomingTrack     chan trackRecvPair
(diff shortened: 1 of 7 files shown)


aler9:
Hello @andresuribe87, i checked the client-side implementation of WebRTC (that is supposedly following the specification) and also the server-side implementation (that was my main doubt) and i can confirm that WebRTC connections can be automatically recovered when `iceConnectionState` is `disconnected`, while they must be recreated when `iceConnectionState` is `failed`, as you mentioned.

With respect to this PR, the check on `iceConnectionState` has to be changed on both client and server, furthermore i recently moved the client-side implementation into a [separate file](https://github.com/bluenviron/mediamtx/blob/main/internal/servers/webrtc/reader.js) - so i'm adapting it in order to account for these changes.

aler9:
merged, thanks for providing the hint!

PR #3927, opened 2024-11-01
fix: improve unsafe path handling

Arcitec:
Thank you so incredibly much for creating MediaMTX! It's practically the perfect media server! I'd like to give back by fixing a few bugs and issues here. :)

**fix: improve unsafe path handling:**

Paths containing spaces or dashes were being interpreted as separate options, since the path handling lacked double quotes.

This fixes all unsafe instances of "PWD" and "HOME", along with all other unsafe paths in the scripts.

**readme: explicitly mount the configuration as read-only:**

This clearly shows users that the MediaMTX container will not modify the configuration file.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1260,7 +1260,7 @@ There are 3 ways to change the configuration:
    * available in the root folder of the Docker image (`/mediamtx.yml`); it can be overridden in this way:
 
      ```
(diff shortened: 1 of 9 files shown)


Arcitec:
Requesting review: @aler9 :)

PR #3941, opened 2024-11-06
Feat: Add additional information in SRT metrics label

pkmdev:
Added stream path and connection address to SRT metrics labels


aler9:
replaced by #4794, that adds path and remoteAddr to all protocols.

Issue #3980, opened 2024-11-26
Windows cmd's hang/don't close on hooks


yelodevopsi:
### Which version are you using?

v1.9.3

### Which operating system are you using?

Windows amd64 standard

### Describe how to replicate the issue

1. Setup the mediamtx.yml with a hook-stream like e.g:

```
paths:
  TEST:
    runOnDemand: gst-launch-1.0 videotestsrc is-live=true ! x264enc tune=zerolatency key-int-max=30 ! rtspclientsink name=sink location=rtsp://localhost:$RTSP_PORT/$MTX_PATH
```    
    
2. Play it with http://localhost:8889/TEST
3. Stop it. See the terminal output that the process is not closed. (gst-launch-1.0 is still running).

**Note:**
PR coming. I've already found a fix into cmd_win.go

### Server logs

```
2024/11/26 08:53:27 INF [WebRTC] [session b7f9df6b] peer connection established, local candidate: host/udp/10.249.0.223/8189, remote candidate: prflx/udp/10.249.0.223/55204
2024/11/26 08:53:27 INF [WebRTC] [session b7f9df6b] is reading from path 'GST-04', 1 track (H264)
Redistribute latency...
2024/11/26 08:58:16 INF [WebRTC] [session b7f9df6b] closed: terminated
2024/11/26 08:58:26 INF [path GST-04] runOnDemand command stopped: not needed by anyone
```

aler9:
fixed by #3981

PR #3981, opened 2024-11-26
Bugfix Closes issue: 3980: Rewrote addProcessToGroup.

yelodevopsi:
Fixed possible invalid process handle by using windows.OpenProcess instead. Fixes Issue https://github.com/bluenviron/mediamtx/issues/3980


aler9:
Thank you very much for this patch, it has been merged. Sorry if it took a while but i had to setup the specific environment (Windows + GStreamer for Windows) and replicate everything.

Issue #4011, opened 2024-12-03
Support embedding mediamtx in other Go binaries


iameli:
### Describe the feature

Hi! I'm working on a video project that, for various reasons, is required to ship as a single statically-linked Go binary. I'd love to use MediaMTX (by forking and calling `core.New`) but as things currently stand I'd have to package up MediaMTX adjacent to my binary. As far as I can tell, there are only two things standing in the way of this being possible:

1. All of MediaMTX's code is in an `internal` subdirectory. I assume that means at some point somebody thought it was a bad idea to support MediaMTX in other Go binaries, so we can have a discussion about that.
2. There's a `go generate` step to create a VERSION file and download HLS.js and mediamtx-rpicamera. There probably exist workarounds for this, such as running this step as part of the process of publishing new package versions and including them in the tagged releases.

Would love to hear your thoughts!

aler9:
Hello,

> All of MediaMTX's code is in an internal subdirectory. I assume that means at some point somebody thought it was a bad idea to support MediaMTX in other Go binaries, so we can have a discussion about that.

This was already discussed in #533. We won't expose a public interface since that would mean guaranteeing its backward compatibility. Just fork the server and move what you need into the `pkg` folder.

> There's a go generate step to create a VERSION file and download HLS.js and mediamtx-rpicamera. There probably exist workarounds for this, such as running this step as part of the process of publishing new package versions and including them in the tagged releases.

We need to run a pre-compilation process as lots of project do. While some compilers allow to define pre-processing stages that are executed automatically (like Meson and Maven), the Go compiler requires a manual command, that is `go generate`, that has to be run before `go build`. Running it does not influence the ability to embed the server into another executable (after you have moved what you need into the `pkg` folder).

PR #4020, opened 2024-12-05
main: expose programmatic entrypoint

iameli:
Per #4011 - I absolutely get not wanting to support a stable Go interface for most things, but would you consider something like this? The only interface this exposes is the command-line interface via a `[]string`, not exactly something that will require a lot of refactoring. But it would still allow embedding MediaMTX elsewhere without entirely forking the repo.

### DIFF
--- a/main.go
+++ b/main.go
@@ -4,13 +4,12 @@ package main
 import (
 	"os"
 
-	"github.com/bluenviron/mediamtx/internal/core"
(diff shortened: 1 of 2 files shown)


aler9:
Sorry but this won't be merged for reasons that have already been explained. If you need to expose an additional interface, you are free to fork the server.

PR #4105, opened 2025-01-03
add PlaybackBaseURL config option

### LINKED ISSUE #4104: Custom base URL in playback list
When mediamtx is located behind a reversing proxy, the URLs provided by the playback `list` endpoint are meaningless to the end client.  There should be some way to modify the base URL included in these responses.

Note @andreymal's comment: https://github.com/bluenviron/mediamtx/issues/3649#issuecomment-2333765100

ijessen-mitll:
Resolves #4104 in a simplistic way.  I would still encourage a generalized mechanism for assigning base/root URLs for each of the various services within mediamtx, but for starters this is functional for our use case.

Note - I did not update the example config file (`mediamtx.yml` ).  The default value preserves existing behavior.

### DIFF
--- a/internal/conf/conf.go
+++ b/internal/conf/conf.go
@@ -209,6 +209,7 @@ type Conf struct {
 	// Playback
 	Playback               bool       `json:"playback"`
 	PlaybackAddress        string     `json:"playbackAddress"`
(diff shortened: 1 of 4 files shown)


aler9:
https://github.com/bluenviron/mediamtx/issues/4104#issuecomment-2953751653

PR #4121, opened 2025-01-08
on connection close, force the immediate session close (triggering the Unread hook).

ftaiolivista:
on connection close, force the immediate session close (triggering the Unread hook).


aler9:
The aim of separating connections from sessions is exactly to provide the ability to detach sessions from connections. Furthermore there are several clients that create a session, close the connection and expect the session to continue working.

PR #4127, opened 2025-01-10
Added "dockerimg"

tranzmatt:
I added a "dockerimg" to allow someone to build a local version of the dockerized mediamtx.  It uses a PLATFORM and USE_FFMPEG make variable setup.  It also installs yt-dlp to allow streaming of some YouTube videos.  I needed it for my purposes, but I suppose it could be removed.


tranzmatt:
Found some bugs, not ready for PR

PR #4168, opened 2025-01-17
rtmp: Support Opus writes from internal stream.

j0sh:
This allows for Opus to go from eg, WHIP -> RTMP

Seems to work OK with recent ffmpeg (master branch, as ffmpeg E-RTMP is still unreleased)


aler9:
Thanks for providing this patch, i've integrated it into the larger #4954, that also checks whether a client supports reading Opus.

PR #4173, opened 2025-01-19
Fix FindSegments when given a start time earlier than all available segments

### LINKED ISSUE #4164: Filtering Timespans returns 0 when starting before any recordings

1.11.0

macOS amd64 Docker

1. start the server
2. Record some segments
3. List recording segments but set the startTime to BEFORE there are any segments `http://10.10.10.89:9996/list?path=4f3b9762-967a-4ff2-a085-ec207f3dafb5&start=2025-01-16T15%3A33%3A00.000Z&end=2025-01-16T18%3A33%3A00.000Z` 

That start time is several hours before i have any recordings for that day. It returns "no recording segments found", even though there are recordings in that timespan, they just are AFTER the start time.

Thank you guys, i love this feature! Just a weird edge case i imagine.

### Server logs

I dont think logs would be helpful in this case, nothing to note there

### Network dump

nothing to note

ijessen-mitll:
fixes #4164 

Could probably use a test.  Sorry about the cruft of the messed up PR previously...

### DIFF
--- a/internal/playback/on_list_test.go
+++ b/internal/playback/on_list_test.go
@@ -19,13 +19,14 @@ import (
 	"github.com/stretchr/testify/require"
 )
 
-func TestOnListUnfiltered(t *testing.T) {
+func TestOnList(t *testing.T) {
 	for _, ca := range []string{
 		"unfiltered",
 		"filtered",
 		"filtered and gap",
 		"different init",
 		"start after duration",
+		"start before first",
 	} {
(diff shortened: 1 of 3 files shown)


aler9:
i added tests and merged, thanks

cooperbrown9:
Hey any idea when the docker image for this will get built? The latest tag is 4 days ago.
Thanks for the quick fix

cooperbrown9:
Or do you have any idea when this might be added to a release?

Issue #1209, opened 2022-10-28
Cache H264 GOPs in order to allow readers to receive frames immediately


aler9:
## Describe the feature

Discussed here #1132

sourcez:
This would be a very desired feature for the video platform we are currently building. Any outlook on when this could be build in? Thanks !

jordybsecurite:
Can you maybe give some pointers as to what you think is necessary to implement this (e.g. which files to edit)? Maybe I can give it a go

aler9:
Hello, in my opinion to implement this you can hack `formatprocessor.H264`, which is an object that is initialized once for every H264 track of every stream:

https://github.com/bluenviron/mediamtx/blob/main/internal/formatprocessor/h264.go

In this structure you have to save all NALUs in a buffer until a key frame is received; when a key frame is received, this buffer must be emptied, and filled with new NALUs starting from the key frame.

When a new RTSP, RTMP of WebRTC client connects (not HLS, since key frame caching is already present in HLS), you have to send to this client all NALUs that are present in the buffer.

0x5e:
want this feature to! I'm using video doorbell with this project, when doorbell event triggered, video showed up in about 3-4s. a gop cache will make the stream open in realtime.

thanks everyone in advance!!

ethaniel:
I have a live stream, where I have multiple webcams in one single location (a zoo). I switch between them every 20 seconds. This feature would help make the switch quick and smooth.

reloadman:
In my opinion, this is a very useful feature that will allow the mediamtx to be used in video surveillance projects. I'm eagerly awaiting its implementation!

jean343:
Thank you for highlighting this issue! We’re actively working on a fix and have successfully gotten the RTSP path working so far. We’re continuing to refine the solution and will contribute the fix to the repository once it’s ready.

We appreciate your patience and look forward to improving this together!

PR #4189, opened 2025-01-23
Cache H264/H265 GOPs in order to allow readers to decode frames immediately

jean343:
### GOP Cache

This PR introduces Group of Pictures (GOP) caching to MediaMTX, enhancing its performance and reducing latency in streaming scenarios. By caching the last GOP for each stream, new subscribers can immediately receive the latest video data without waiting for the next keyframe, improving the user experience, especially for streams with long keyframe intervals.

This works for both H264 and H265, as well as for RTSP and WebRTC.

### Configurable Cache Settings:
Introduced a new configuration parameter gopCache in mediamtx.yml for enabling/disabling GOP caching.

Fix: https://github.com/bluenviron/mediamtx/issues/1209

### DIFF
--- a/internal/stream/stream_format.go
+++ b/internal/stream/stream_format.go
@@ -6,11 +6,17 @@ import (
 
 	"github.com/bluenviron/gortsplib/v4/pkg/description"
 	"github.com/bluenviron/gortsplib/v4/pkg/format"
+	"github.com/bluenviron/mediacommon/pkg/codecs/h264"
+	"github.com/bluenviron/mediacommon/pkg/codecs/h265"
 	"github.com/pion/rtp"
 
 	"github.com/bluenviron/mediamtx/internal/formatprocessor"
 	"github.com/bluenviron/mediamtx/internal/logger"
 	"github.com/bluenviron/mediamtx/internal/unit"
+)
+
+const (
+	maxCachedGOPSize int = 512
 )
 
 func unitSize(u unit.Unit) uint64 {
@@ -19,6 +25,26 @@ func unitSize(u unit.Unit) uint64 {
 		n += uint64(pkt.MarshalSize())
 	}
 	return n
+}
+
+func isKeyFrame(u unit.Unit) bool {
+	switch tunit := u.(type) {
+	case *unit.H264:
+		return h264.IDRPresent(tunit.AU)
+	case *unit.H265:
+		return h265.IsRandomAccess(tunit.AU)
+	}
+	return false
+}
+
+func isEmptyAU(u unit.Unit) bool {
+	switch tunit := u.(type) {
+	case *unit.H264:
+		return len(tunit.AU) == 0
+	case *unit.H265:
+		return len(tunit.AU) == 0
+	}
+	return true
 }
 
 type streamFormat struct {
@@ -30,6 +56,7 @@ type streamFormat struct {
 	proc           formatprocessor.Processor
 	pausedReaders  map[*streamReader]ReadFunc
 	runningReaders map[*streamReader]ReadFunc
+	gopCache       bool
 }
 
 func (sf *streamFormat) initialize() error {
@@ -78,7 +105,7 @@ func (sf *streamFormat) writeRTPPacket(
 	ntp time.Time,
 	pts int64,
 ) {
-	hasNonRTSPReaders := len(sf.pausedReaders) > 0 || len(sf.runningReaders) > 0
+	hasNonRTSPReaders := len(sf.pausedReaders) > 0 || len(sf.runningReaders) > 0 || sf.gopCache
 
 	u, err := sf.proc.ProcessRTPPacket(pkt, ntp, pts, hasNonRTSPReaders)
 	if err != nil {
@@ -93,6 +120,32 @@ func (sf *streamFormat) writeUnitInner(s *Stream, medi *description.Media, u uni
 	size := unitSize(u)
 
 	atomic.AddUint64(s.bytesReceived, size)
+
+	if sf.gopCache && medi.Type == description.MediaTypeVideo {
+		if isKeyFrame(u) {
+			if s.CachedUnits == nil {
+				// Initialize the cache and enable caching
+				s.CachedUnits = make([]unit.Unit, 0, maxCachedGOPSize)
+			} else {
+				// Keep the last packets that were used to generate the key frame.
+				// This is to send a full key frame in the RTSP stream.
+				i := len(s.CachedUnits)
+				for ; i > 0; i-- {
+					if !isEmptyAU(s.CachedUnits[i-1]) {
+						break
+					}
+				}
+				s.CachedUnits = s.CachedUnits[i:]
+			}
+		}
+		if s.CachedUnits != nil {
+			s.CachedUnits = append(s.CachedUnits, u)
+		}
+		l := len(s.CachedUnits)
+		if l > maxCachedGOPSize {
+			s.CachedUnits = s.CachedUnits[l-maxCachedGOPSize:]
+		}
+	}
 
 	if s.rtspStream != nil {
 		for _, pkt := range u.GetRTPPackets() {
--- a/mediamtx.yml
+++ b/mediamtx.yml
@@ -23,6 +23,9 @@ writeQueueSize: 512
 # Maximum size of outgoing UDP packets.
 # This can be decreased to avoid fragmentation on networks with a low UDP MTU.
 udpMaxPayloadSize: 1472
+# Enable GOP cache to improve initial playback experience for new clients.
+# Note: will increase memory usage.
+gopCache: false
 
 # Command to run when a client connects to the server.
 # This is terminated with SIGINT when a client disconnects from the server.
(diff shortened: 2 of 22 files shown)


aler9:
This is a great work. The important thing is that you can confirm that the things works and in which scenarios (protocols, codecs, with/without B-frames).
We can then adjust details, the ones that come into my mind are:
- check on the GOP size to prevent RAM exhaustion
- additional codecs, at least AV1

jean343:
Thanks @aler9  for your kind comment!

The feature does work in the following scenarios:

Protocol:

- WebRTC
- RTSP

Codecs:

- H.264
- H.265

I have only tried videos without b-frames as WebRTC does not support b-frames. There might be adjustments to make when dealing with b-frames over RTSP.

In order to reduce RAM exhaustion, we do not cache anything until we get a key frame, this will prevent unsupported codecs from storing anything, and will save a little bit for supported codecs.
In case the GOP is really long, we trim at `512 packets`, conserving memory. In this case, clients will need to wait until the next key frame before video playback.

As per additional codecs, I could not find a reliable way to detect their keyframes.

yairzahavi:
You have missing fixes needed in order to align with base branch.
I did the in my own fork since i needed the Gop Cache and base branch alignment..

you can reference them here:
https://github.com/bluenviron/mediamtx/pull/4282

angry-beaver:
@aler9 @jean343 This is a cool feature. Do you have any thoughts on when it will be completed?

aler9:
@angry-beaver i need to finish a couple of other things then i'll focus on this. In the meanwhile, @jean343 and @yairzahavi can try to go on by themselves.

yairzahavi:
> @angry-beaver i need to finish a couple of other things then i'll focus on this. In the meanwhile, @jean343 and @yairzahavi can try to go on by themselves.

@jean343 I'll try and do it this soon including the av1 and ram exhaustion prevention and i'll ping you for a review. 🙌

jean343:
Thanks everyone for the help. I will merge from master and fix the build.

The AV1 work is awesome.

I did intentionally keep the cached packets of the last key frame. They are needed if a user connects with RTSP.
Try with ffplay:
`ffplay rtsp://localhost/streamid`

internal/stream/stream_format.go
```diff
@@ -94,6 +128,33 @@ func (sf *streamFormat) writeUnitInner(s *Stream, medi *description.Media, u uni
...
+		}
+		if s.CachedUnits != nil {
+			s.CachedUnits = append(s.CachedUnits, u)
+		}
+		l := len(s.CachedUnits)
+		if l > maxCachedGOPSize {
+			s.CachedUnits = s.CachedUnits[l-maxCachedGOPSize:]
+			sf.decodeErrLogger.Log(logger.Warn, "GOP cache is full, dropping packets")
+		}
+	}
```
yairzahavi, line 156: @jean343
> Thanks everyone for the help. I merged from master and fixed the build!
> 
> @yairzahavi, the AV1 work is awesome. I merged the AV1 work into this branch and fixed the merge conflicts. I did not change the CacheLength logic, it it's not clear that the additional logic helps performance.
> 

You are right that the change is not clear and additionally it doesn't really work and i have yet to figure out why.
 
But the reason I tried and change it is that i inspected this code block and it seems there is another memory allocation when you surpass the maxCachedGOPSize
And if you truncate it afterwards the memory allocation already happened.
Additionally I tried to reduce memory allocations and usage by allocating a fixed size.

I hope you'd have a solution\idea for this.

yairzahavi:
Another point i noticed is that the GOP cache causes higher webrtc jitter
So i'm contemplating how and if to solve this as well.

aler9:
Hello, i've tested the patch, while the working principle is present, there some aspects that can be improved:

1. the startup phase is not pleasant as all past frames of the GOP are shown and played very fast (see attached video). I understand that this behavior is needed as workaround to prevent freezing, but there must be some alternative. For instance, we had a similar problem when implementing the Playback server. In that case, duration of past frames is set to zero, because the transport mean (which is MP4) allows that (more or less!). However i know that it's not possible to do that with most streaming protocols. An alternative might be grouping multiple H264 access units into a big H264 access unit that contains all frames.
2. audio sync: linked to point 1. If you add or remove some delta T from video, then audio will get desynced.
3. RTSP: this patch does not cover RTSP. In RTSP, packets are not written to individual readers, but they are sent by calling  `ServerStream.WritePacketRTP` once, in order to support the multicast transport, in which packets are sent once to the network.
   There should be a mechanism in which when a RTSP reader is created and the transport protocol is not Multicast, then GOP frames are sent through `ServerSession.WritePacketRTP`.

The GOP caching feature has always been difficult to implement because it has to take into consideration how players react when receiving a bunch of access units at the same time. It involves testing all possible ways to send the GOP, digging into source code of all players and codec specifications.

The feature can be merged into the main branch only when a high level of compatibility with all major protocols and players is reached.

https://github.com/user-attachments/assets/c7247a35-689e-42de-b840-ec86e389462e

internal/stream/stream_format.go
jean343: We could drop the entire cache once it reaches `maxCachedGOPSize`. It would solve memory allocations, and once the cache size reaches `maxCachedGOPSize` the GOP gets affected and the video player will need to wait for the next key frame regardless.

jean343:
Thank you @aler9 for testing and for your feedback.

I did not expect to uncover this many corner cases when I started implementation :)

1. I agree that playing back all frames very quickly is not optimal, in our testing, where our GOP setting is ~12s, it ends up being better than waiting up to 12s for video playback.
As you know, we can't set the duration to 0s, and because the transport is UDP, we can not send an unlimited amount of packets instantly.
We have tested sending a key frame, and dropping all p-frames until the next key frame. In this case, playback starts instantly, and obviously freezes until next key frame. This indicates that we could possibly combine the p-frames into one...
It would be amazing if you could help merging the frames together into one!
2. I do not believe audio desync should be a problem, because we do not play audio back during catch up.
3. Unicast RTSP is supported, and in this case, playing back all frames at the same time does work, which is great.
Multicast RTSP should not have GOP cache enabled, is there a check in the code I could add to disable that portion?

internal/stream/stream_format.go
```diff
@@ -94,6 +128,33 @@ func (sf *streamFormat) writeUnitInner(s *Stream, medi *description.Media, u uni
 
 	atomic.AddUint64(s.bytesReceived, size)
 
+	if sf.gopCache && medi.Type == description.MediaTypeVideo {
+		if isKeyFrame(u) {
```
HugeLeaf, line 132: would it be better to record from SPS/PPS?

mediamtx.yml
```diff
@@ -23,6 +23,9 @@ writeQueueSize: 512
 # Maximum size of outgoing UDP packets.
 # This can be decreased to avoid fragmentation on networks with a low UDP MTU.
 udpMaxPayloadSize: 1472
+# Enable GOP cache to improve initial playback experience for new clients.
+# Note: will increase memory usage.
+gopCache: false
```
JeDiE99, line 28: maybe need parameter(in bytes ? per path ?) for more transparent memory control ?

jean343: The cache is in-memory and does not need a path. I like the idea of specifying the cache size, in bytes or packets.

tuan3w:
It's interest to see this feature is merged ? Any progress here. Thanks

jean343:
> It's interest to see this feature is merged ? Any progress here. Thanks

I would definitely like to see this being merged.

I can not find a good solution for issue 1. in this comment. If we accelerate playback too much, client refuses to load video.
https://github.com/bluenviron/mediamtx/pull/4189#issuecomment-2727376403

Maybe we could indicate it in the docs, and leave feature disabled by default.

Javier-d98:
This feature would be great for me too
I think the current implementation is good enough, specially if it's disabled by default as @jean343 suggests.

xiaoxuan010:
> 1. I agree that playing back all frames very quickly is not optimal, in our testing, where our GOP setting is ~12s, it ends up being better than waiting up to 12s for video playback.
>    As you know, we can't set the duration to 0s, and because the transport is UDP, we can not send an unlimited amount of packets instantly.
>    We have tested sending a key frame, and dropping all p-frames until the next key frame. In this case, playback starts instantly, and obviously freezes until next key frame. This indicates that we could possibly combine the p-frames into one...
>    It would be amazing if you could help merging the frames together into one!

Thanks for your contribution! It seems that handling the _previous frames_ (that occurs before the new reader comes into the server) is a big problem? The point of _combining the p-frames_ reminds me, and I found a [paper](https://www.techscience.com/cmc/v86n1/64408) that implemented the feature in a very similar and effective way!

In short, the article introduces a video decoder and a temp video encoder on the server side, always providing the newcomers with the newest frames (which are re-encoded from the normal stream, to a temp GOP that starts with a I-frame). I believe that this mechanism is effective and elegant, while the only problem is that it needs a kind of FFmpeg stuff on the server side, which I'm not sure whether we could implement by kind like [external command hook](https://mediamtx.org/docs/usage/hooks), other than embedding the decoder and encoder inside MediaMTX.

Refs:
- [EGOP: A Server-Side Enhanced Architecture to Eliminate End-to-End Latency Caused by GOP Length in Live Streaming](https://www.techscience.com/cmc/v86n1/64408)
- [EGOP stream server](https://github.com/zhoukunpeng504/EGOP-stream-server)

PR #4218, opened 2025-02-02
Add FreeBSD instructions

diizzyy:
Committed to ports tree as of https://github.com/freebsd/freebsd-ports/commit/981b361efddf111d3ed5def597ce365672ac4a9e


aler9:
i see that the porting is maintained and compiled from source so i think it can be added to the README. Thanks for this work!

PR #4252, opened 2025-02-13
Update GitHub workflows for faster builds using a matrix strategy

DanTheMan827:
Refactor the GitHub Actions workflows to utilize a matrix for building binaries across multiple platforms, improving build efficiency.


aler9:
Sorry but we can't merge this. Although more efficient on GitHub, we also need to use the standalone `make binaries` for internal reasons.

PR #4282, opened 2025-02-26
Feature/mediamtx gop cache

yairzahavi:
a fork of https://github.com/bluenviron/mediamtx/pull/4189

> GOP Cache
> This PR introduces Group of Pictures (GOP) caching to MediaMTX, enhancing its performance and reducing latency in streaming scenarios. By caching the last GOP for each stream, new subscribers can immediately receive the latest video data without waiting for the next keyframe, improving the user experience, especially for streams with long keyframe intervals.
> 
> This works for both H264 and H265, as well as for RTSP and WebRTC.
> 
> Configurable Cache Settings:
> Introduced a new configuration parameter gopCache in mediamtx.yml for enabling/disabling GOP caching.
> 
> Fix: https://github.com/bluenviron/mediamtx/issues/1209 @jean343 

> The feature does work in the following scenarios:
> 
> Protocol:
> 
> * WebRTC
> * RTSP
> 
> Codecs:
> 
> * H.264
> * H.265
> 
> I have only tried videos without b-frames as WebRTC does not support b-frames. There might be adjustments to make when dealing with b-frames over RTSP.
> 
> In order to reduce RAM exhaustion, we do not cache anything until we get a key frame, this will prevent unsupported codecs from storing anything, and will save a little bit for supported codecs. In case the GOP is really long, we trim at `512 packets`, conserving memory. In this case, clients will need to wait until the next key frame before video playback.
> 
> As per additional codecs, I could not find a reliable way to detect their keyframes.
> 
> In the WebRTC playback scenario, the PTS and Timestamp needs to be modified to prevent gaps. WebRTC will pause and stop playback if set incorrectly.
> 
> For example, incorrect Timestamp will look like:
> 
>  Untitled.mov 
> Correct timestamp will look like:
> 
>  Screen.Recording.2025-01-23.at.1.40.16.PM.mov

[ ] Fixed Compatibility with mediacommon v2
[ ] Added Support for AV1

### DIFF
--- a/apidocs/openapi.yaml
+++ b/apidocs/openapi.yaml
@@ -71,6 +71,8 @@ components:
           type: boolean
         runOnDisconnect:
           type: string
+        gopCache:
+          type: boolean
 
         # Authentication
         authMethod:
--- a/internal/conf/conf.go
+++ b/internal/conf/conf.go
@@ -169,6 +169,7 @@ type Conf struct {
 	RunOnConnect        string          `json:"runOnConnect"`
 	RunOnConnectRestart bool            `json:"runOnConnectRestart"`
 	RunOnDisconnect     string          `json:"runOnDisconnect"`
+	GopCache            bool            `json:"gopCache"`
 
 	// Authentication
 	AuthMethod                AuthMethod                  `json:"authMethod"`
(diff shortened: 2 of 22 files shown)


yairzahavi:
hey @jean343  @aler9.
I'd be happy if the both of you could take a look on my PR.
I changed @jean343 caching logic to try and limit cache usage and memory allocation as much as possible yet i'm not confident as i never worked with go or video encodings before today.

PR #4324, opened 2025-03-09
Docker RPi without privileged

nvtkaszpir:
Adding minimal working example docker-compose.yml to the README.md to help running mediamtx in a bit more secure way via docker on Raspberry Pi 4 with official RPi CSI camera.
Tested under Debian 12 Bookworm kernel 6.6.62+rpt-rpi-v8 aarch64 and Docker Community edition version 28.0.1.

Explanation line by line of the specific lines.

* Use official docker image for rpi with ffmpeg.
* Switch to user nobody:video which is present in the image, this allows to access camera devices defined later.
* Set read only mode on the whole container, from now on we must explicitly define writeable mount points.
* Expose required ports.
* Set env vars for RTSP tcp mode, pass ip address
* Mount volumes as read only with an app config and udev.
* Mount tmpfs with exec mode under /dev/shm where mediamtx creates resources.
* Pass devices to the container:
  - v4l subsystem
  - dma
  - pass on all devices which are used by mediamtx to access CSI camera - video and media

### DIFF
--- a/README.md
+++ b/README.md
@@ -638,6 +638,64 @@ paths:
 
 All available parameters are listed in the [sample configuration file](/mediamtx.yml).
 
+If you wish to run mediamtx without `privileged` and with read only filesystem 
+in a container then you must pass tmpfs with exec and list of devices
+which are specific to the given camera.
+
+Below is a minimal example docker-compose.yml for standard Raspberry Pi CSI camera
+
+```yml
+---
+# run docker compose up on rpi4 with CSI camera on Debian Bookworm (12)
+services:
+  mediamtx:
+    image: "docker.io/bluenviron/mediamtx:latest-ffmpeg-rpi"
+    user: "65534:44" # nobody:video inside of the container
+    read_only: true
+    ports:
+      - "8554:8554"
+      - "1935:1935"
+      - "8888:8888"
+      - "8889:8889"
+      - "8890:8890/udp"
+      - "8189:8189/udp"
+    environment:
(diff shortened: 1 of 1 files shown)


aler9:
Hello, thanks for providing an alternative to the `--privileged` flag for interacting with the RPI Camera.

However, you're are also introducing an additional overlay, which is Docker Compose, which is something that some users might not want to add to their stack for several reasons. The aim of the README is to provide examples and instructions that require as few components as possible.

Therefore, it would be better if you move all the additional devices, volumes and user ID into the existing Docker command in the README, in place of `--privileged`.

nvtkaszpir:
OK, I'll try to test it later and in that case I believe I'd rather make a new PR instead of altering this one.

Why to keep this PR?
First of all with plain docker command it becomes a really long command with so many devices to pass on into the container, thus it becomes a bit problematic for some and that's why I prefer docker-compose way (especially if you tend to keep configuration in the repo). Thus I think it would be beneficial to create something like a `hack/mediamtx/rpi-csi/` directory with `docker-compose.yaml` file in it and move that code there, which would be useful minimal reference for the others. Of course this may beviewed as unwanted bloat/clutter in the repo. 

So what is the view in doing `hack/` directory? If it is not going to be approved then this PR can be closed and I'll just make a new one with the pure docker cli command options only.

nvtkaszpir:
As requested, providing PR with docker cli command only https://github.com/bluenviron/mediamtx/pull/4326

aler9:
replaced by #4347

PR #4326, opened 2025-03-09
Add docker RPi cam without elevated privileges

nvtkaszpir:
Add example command to run docker container on Raspberry Pi with CSI camera without needing privileged mode.
Providing read only root filesystem, user nobody:video to access devices, pass down camera from the host.

Network host was preserved to decrease clutter caused by additional port forwards.

Tested under Debian 12 Bookworm kernel 6.6.62+rpt-rpi-v8 aarch64 and Docker Community edition version 28.0.1, with default config from the within the container.

### DIFF
--- a/README.md
+++ b/README.md
@@ -638,6 +638,45 @@ paths:
 
 All available parameters are listed in the [sample configuration file](/mediamtx.yml).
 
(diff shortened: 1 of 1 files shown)


aler9:
replaced by #4347

PR #4347, opened 2025-03-16
add instructions on using the RPI Camera without --privileged (#4326) (#4324)

aler9:


aler9:
turned out that this cannot be merged, since the list of `/dev/video*` and `/dev/media*` devices varies depending on the specific RPI model.

PR #4356, opened 2025-03-19
add prefix for syslog

JeDiE99:
view sys log for several processes


aler9:
everything seems to be in order - this has been merged, thanks!

PR #4357, opened 2025-03-20
[chore]: Bugfix: removed duplicate paths_* and hls_muxers_* patterns

yelodevopsi:
* Rewrite to pattern array: This gives specific error to which line fails in test32 and test64.
* The old code just tells "something" in the regexp is wrong, but not which

### DIFF
--- a/internal/core/metrics_test.go
+++ b/internal/core/metrics_test.go
@@ -319,128 +319,123 @@ webrtc_sessions_bytes_sent 0
 
 		bo := httpPullFile(t, hc, "http://localhost:9998/metrics")
 
-		require.Regexp(t,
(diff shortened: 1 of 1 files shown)


aler9:
Hello, this is nice but there are a couple of missing things:

* the new test does not check for the presence of additional lines other than the ones defined inside `patterns`. For instance, if there's a extraneous metric called `webrtc_should_not_be_here`, the test allows it (while it shouldn't)
* the new test does not check for the presence of repeated lines - for instance, if there are two metrics `srt_conns_mbps_send_rate` and `srt_conns_mbps_send_rate`, the test allows it (while it shouldn't)

can you upgrade the code in order to account for these missing test cases?

aler9:
closed since remarks were not addressed.

Issue #4388, opened 2025-04-01
Allow support for MPEG-TS over unix sockets


jean343:
#### **Background**
We are operating a high-scale video ingestion pipeline where **hundreds of MPEG-TS feeds** are streamed into **MediaMTX** for real-time processing and distribution. Currently, we rely on **UDP ports** to deliver these streams.

However, this approach has become increasingly **difficult to manage and maintain** due to:

- **Port exhaustion and conflicts**, especially in environments with dynamically provisioned feeds.
- **Lack of backpressure or error signaling**, which is typical with UDP.
- **Fragility** in containerized or orchestrated environments where port mappings and host-level networking complicate deployment.

#### **Proposal**
We request **support for receiving MPEG-TS streams over Unix domain sockets** as a new input transport option in MediaMTX.

#### **Benefits**
- **Simplified Port Management**: Eliminate the need to manage hundreds of distinct UDP ports. Each feed can be associated with a well-named socket file.
- **Improved Reliability**: Unix sockets offer **connection-oriented semantics**, improved **error detection**, and **automatic cleanup** on disconnection.
- **Better Integration with Local Producers**: For tightly coupled pipelines (e.g., processes on the same host), Unix sockets provide **low-latency** and **secure** IPC.
- **Enhanced Deployment Flexibility**: Especially in **Docker/Kubernetes** environments, volume-mounted socket paths are more portable and easier to manage than host network bindings.

#### **Use Case**
We envision feed-producing applications piping MPEG-TS data to MediaMTX by connecting to a domain socket (e.g., `/var/run/mediafeeds/camera123.sock`). MediaMTX would treat this as a live input stream, similarly to a UDP listener, but initiated over a **stream-oriented** UDS.

RedShift1:
Was this feature request written by AI?

jean343:
It's not an AI feature request, we actually do need this feature to reliably generate streams at runtime. The alternative is to use UDP ports, and this causes all sort of port conflicting issues.

I might have asked AI to polish my feature request :)

RedShift1:
Suspected as much, it's a useful feature but parts of the rationale don't add up. UNIX sockets don't deal with backpressure and error signaling, just like UDP sockets. It also doesn't have any connection-oriented semantics, error detection or automatic cleanup. It's really just a dumb pipe like an UDP socket with the only difference it doesn't involve the network stack. But the semantics of it are completely determined by the applications reading and writing to the UNIX socket.

PR #4389, opened 2025-04-01
feat: add support for MPEG-TS over unix

jean343:
fixes: #4388

# For testing

Settings:
```
paths:
  ffmpeg:
    source: unix:/tmp/ffmpeg.socket
```

Populate unix socket:
```
ffmpeg -re -f lavfi -i testsrc=size=1280x720:rate=30 \
-c:v libx264 -pix_fmt yuv420p -preset ultrafast -b:v 600k \
-f mpegts "unix:/tmp/ffmpeg.socket"
```

Play the video:
```
ffplay -i rtsp://localhost:8554/ffmpeg
```


aler9:
while the feature is useful, this implementation lacks some details:

* it doesn't provide deadlines
* it creates an additional static source instead of using the existing UDP one, although the application layer is the same

I wrote an alternative implementation here: #4828

Issue #4319, opened 2025-03-07
Get a frame


pikachu937:
### Describe the feature

Sometimes it is necessary to get not a video, but a frame and specifically by means of the server without using ffmpeg/gstreamer/opencv and the like.
For example, a preview image from a stream or to see what is happening in front of the camera at the moment, for example through a call to the api.
Is this functionality planned?

aler9:
Sorry but this can't be implemented since it would require a video decoder, that we are not going to add to the server for practical and legal reasons.

PR #4433, opened 2025-04-14
Feat/get frame

galindocode:


galindocode:
@aler9 i'm working in #4319 and i'm think which is the best server for include this new endpoint.

Currently i'm adding to playback server. What do you think?

aler9:
Hello, this can't be merged for several reasons:

* The H264 decoding process is protected by active patents, therefore there are legal implications in embedding FFmpeg into the server.
* FFmpeg bindings are compatible with the FFmpeg version that was used during compilation only, therefore it would be impossible to produce universal binaries like we are doing right now.
* FFmpeg bindings require FFmpeg to be installed.

I'd really like to have this feature in the future, but for the moment it would cause more harm than good. Nothing prevents you from building a platform in which you have MediaMTX on one side, and a microservice on the other that is in charge of periodically connecting to MediaMTX and generate thumbnails.

PR #4460, opened 2025-04-22
Add authentication method internal-and-jwt

kalwinskidawid:
Added authentication method: `internal-and-jwt`, which allows authentication by internal user in `mediamtx.yml` and JWT. 

Useful if another service in docker compose would like to, for example, modify the configuration in real time.

### DIFF
--- a/internal/auth/manager.go
+++ b/internal/auth/manager.go
@@ -132,8 +132,14 @@ func (m *Manager) Authenticate(req *Request) error {
 	case conf.AuthMethodHTTP:
 		err = m.authenticateHTTP(req)
 
-	default:
+	case conf.AuthMethodJWT:
 		err = m.authenticateJWT(req)
+
+	default:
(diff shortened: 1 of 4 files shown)


aler9:
Hello, it's already possible to combine JWT and an internal user database by using a local identity server (like Keycloak), federated with the original identity server and with additional user entries.

Adding this into the server would increase the maintenance burden without adding new features and with a very limited user base.

kalwinskidawid:
Fine, **Keycloak** is the solution. If, on the other hand, one does not use Keycloak, but if one only has a JWT authorization server, I don't see the point of switching specifically to Keycloak - because it makes another layer to manage - but I understand your concern about unnecessary solutions in your program. In general, I just meant to have the possibility of a “service” account, which has the ability to communicate through the API and does not need to generate a new JWT every X min, because it also gets messy.

Issue #4452, opened 2025-04-20
Add an API endpoint to rotate the JWKS key


aler9:
### Describe the feature

Discussed in #4451

dan-nicholls:
Have addressed this in #4463, open for feedback.

aler9:
fixed by #4463

PR #4463, opened 2025-04-24
Feat: Add JWKS rotation API endpoint

dan-nicholls:
Hi all, this is my first contribution and should addresses #4452. The only blaring issue is that the endpoint still requires a JWT signed with a key from the outdated JWKS, but more than open to feedback.

Changes include:
- Seperating refresh and pulling of JWKS tokens
- Add `/auth/jwks/refresh` endpoint for force refreshing the JWKS token
- Updated OpenAPI Schema definition

### DIFF
--- a/apidocs/openapi.yaml
+++ b/apidocs/openapi.yaml
@@ -1364,6 +1364,22 @@ paths:
               schema:
                 $ref: '#/components/schemas/Error'
 
+  /v3/auth/jwks/refresh:
+    post:
+      operationId: authJwksRefresh
+      tags: [Authentication]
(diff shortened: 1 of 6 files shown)


aler9:
i added the following:

* there was a race condition that allowed multiple clients to download the same JWKS in parallel, due to the mutex being acquired and released twice. Now the JWKS is downloaded once. During the download, all clients are put on hold. If the download fails, the download is retried by the next client. If the download is successful, all following clients use the same JWKS without downloading it again (unless the JWKS is refreshed or 1 hour has passed).

* i added tests

then i merged. Thanks for the contribution!

PR #4483, opened 2025-05-02
chore(image/rpi): update base images to latest versions for ARM archs

royveshovda:
Bump the Raspberry pi images to the newest versions.

Sources:

- <https://hub.docker.com/layers/balenalib/raspberry-pi/bullseye-run-20250501/images/sha256-415949e4f375c9efec7c2298bddfe0904748c16624a728f4c3a9fa4b0212af62>
- <https://hub.docker.com/layers/balenalib/raspberrypi3-64/bullseye-run-20250401/images/sha256-3aad00a3bf4caa504329e5d8d2a789c88f6136c2422e238f1d78410312c1992c> (newest at the time of PR)

### DIFF
--- a/docker/ffmpeg-rpi.Dockerfile
+++ b/docker/ffmpeg-rpi.Dockerfile
@@ -7,9 +7,9 @@ ADD binaries/mediamtx_*_linux_arm64.tar.gz /linux/arm64
 
 #################################################################
 
-FROM --platform=linux/arm/v6 balenalib/raspberry-pi:bullseye-run-20240508 AS base-arm-v6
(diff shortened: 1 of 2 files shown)


aler9:
Hello, we cannot do this since we are running mediamtx-rpicamera in these images, and mediamtx-rpicamera is linked against libraries contained in these images: https://github.com/bluenviron/mediamtx-rpicamera/blob/main/utils.mk
Maybe upgraded version of these libraries are backward-compatible, maybe they are not.

Therefore: we will definitely upgrade these images in the future, but only when we are forced to do it (usually due to new hardware).

royveshovda:
OK. I would encourage you to consider the security risk this implies, as the images get older and older. Trivy is screaming at me when I am using the latest rpi image

Issue #4276, opened 2025-02-24
No recording segments found asking near a segment start time inside a timespan returned by the


RafaMarchante:
### Which version are you using?

1.10.0

### Which operating system are you using?

Linux amd64 Docker

### Describe how to replicate the issue

When making an HTTP request to the playback server to list the available recordings from a stream, I get a single timespan starting at a given time that has a duration of well over an hour, as you can see in the following image:

![Image](https://github.com/user-attachments/assets/cd45322b-3acc-4644-aabe-ec52770b92eb)

We then try to make a GET request to the playback server to get a video, using a start query parameter that is just milliseconds earlier than one of the recording segments found on the recordings folder where they are stored:

![Image](https://github.com/user-attachments/assets/5e8a052a-7d67-4067-8370-52ed8f6b478d)

As you can observe, the timestamp is inside the timespan returned by the list HTTP request. When we try to make the GET request we get the message "no recording segments found":

![Image](https://github.com/user-attachments/assets/665966e9-7bb5-49d4-b858-530a10aa7d8e)

As you can see, the start query parameter timestamp has a value of 676 milliseconds, and the timestamp found on the recording file name is 687. 

If we increase or decrease the start timestamp requested, then we are able to download the video without any problem, using milliseconds values of 587 or 787.

tvidaldev:
Hi,

It still happens in v1.12.2.
I think I managed to fix it.
Unfortunately I cannot reproduce it all the time. It seems to depend on the MP4 files written themselves.
I will submit a PR later today.

PR #4533, opened 2025-05-16
playback: fix fMP4 segment seek when start offset is in the last entry of trun

tvidaldev:
### Description

This PR is the result of the investigation of the issue [4276](https://github.com/bluenviron/mediamtx/issues/4276).

For example let's say you are recording a source with MediaMTX, and you have on the disk the following files:
- ...
- 2025-05-16_12-27-50-415642.mp4
- 2025-05-16_12-29-50-421186.mp4
- ...

If you try [start=2025-05-16T12:29:50.421186](http://localhost:9996/get?duration=20&path=train_follower_cam&start=2025-05-16T12%3A29%3A50.421186%2B02%3A00) it's working fine.
But if you try 1 microsecond before the start of the MP4 as in the list, [start=2025-05-16T12:29:50.421185](http://localhost:9996/get?duration=20&path=train_follower_cam&start=2025-05-16T12%3A29%3A50.421185%2B02%3A00) you'll receive the error `no recording segments found`.

I cannot reproduce it everytime, but when it fails, applying the fix of this PR solves the problem.

It makes sense because the offset may be present in the last entry of the last `trun` box, and the condition should be check at the end of the loop, not at the beginning.

### DIFF
--- a/internal/playback/segment_fmp4.go
+++ b/internal/playback/segment_fmp4.go
@@ -435,10 +435,6 @@ func segmentFMP4SeekAndMuxParts(
 					break
 				}
 
-				if muxerDTS >= 0 {
-					atLeastOnePartWritten = true
-				}
-
 				sampleOffset := dataOffset
 				sampleSize := e.SampleSize
 
@@ -467,6 +463,10 @@ func segmentFMP4SeekAndMuxParts(
 				atLeastOneSampleWritten = true
 				dataOffset += uint64(e.SampleSize)
 				muxerDTS += int64(e.SampleDuration)
+
+				if muxerDTS >= 0 {
+					atLeastOnePartWritten = true
+				}
 			}
 
 			if atLeastOneSampleWritten {


aler9:
Hello, thanks for providing the analysis and the fix, however the code is not 100% OK, since it allows the playback server to return empty recordings. For instance, let's suppose there's this segment:

```
2008-11-07_11-22-00-000000.mp4
```

Let's suppose it contains a 1 second-long frame:

```go
fmp4.Parts{
		{
			SequenceNumber: 1,
			Tracks: []*fmp4.PartTrack{
				{
					ID: 1,
					Samples: []*fmp4.PartSample{
						{
							Duration:        1 * 90000,
							IsNonSyncSample: false,
							Payload:         []byte{1, 2},
						},
					},
				},
			},
		},
	}
}
```

If we call `get?start=2008-11-07T11:22:00.50`, the server will return code 200 and an MP4 with an empty list.
The server should only return frames with a start date that is equal or greater than start.

I will provide a patch (and unit tests!) later.

aler9:
replaced by #4539

PR #4539, opened 2025-05-18
playback: fix error 404 when seeking before start of segment (#4276) (#4533)

### LINKED ISSUE #4276: No recording segments found asking near a segment start time inside a timespan returned by the

aler9:
Fixes #4276
Replaces #4533

### DIFF
--- a/internal/playback/segment_fmp4.go
+++ b/internal/playback/segment_fmp4.go
@@ -10,7 +10,6 @@ import (
 
 	"github.com/abema/go-mp4"
 	"github.com/bluenviron/mediacommon/v2/pkg/formats/fmp4"
-	"github.com/bluenviron/mediamtx/internal/recordstore"
 )
 
 const (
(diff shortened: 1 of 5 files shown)


tvidaldev:
I can confirm the PR fixes the issue.
Waiting now for the linter issue, anyway I can use the branch fix/playback-between until merged.
Thanks @aler9!

internal/playback/segment_fmp4.go
```diff
@@ -490,10 +484,6 @@ func segmentFMP4SeekAndMuxParts(
 		return 0, err
 	}
 
-	if !atLeastOnePartWritten {
```
IDisposable, line 493: This duplicate code warning is not _exactly_ true... these two `func` are now **almost** identical, except for one line 

In `segmentFMP4SeekAndMuxParts` [line 427](https://github.com/bluenviron/mediamtx/pull/4539/files#diff-17995db508ad7dc9d924e5aff2204e8376fcd5b8c6160eccc191fb25ed1b0fceR427) reads
```go
muxerDTS := int64(tfdt.BaseMediaDecodeTimeV1) - segmentStartOffsetMP4
```

In `segmentFMP4MuxParts` [line 547](https://github.com/bluenviron/mediamtx/pull/4539/files#diff-17995db508ad7dc9d924e5aff2204e8376fcd5b8c6160eccc191fb25ed1b0fceR547) reads
```go
muxerDTS := int64(tfdt.BaseMediaDecodeTimeV1) + segmentStartOffsetMP4
```

I assume that one of these two is incorrect because the first one `-` subtracts the `segmentStartOffsetMP4` while the second one `+` adds the `segmentStartOffsetMP4`.  

I am not sure that's an intentional difference.

If not intentionally different, I am not sure which one is **correct** (I suspect the `+` addition is right... but I have no idea).  After anointing the correct version, all calls to the other can be swapped and the (now entirely) duplicate can be deleted. In other words, pick if it's supposed to be `+` or `-` and replace all calls to `segmentFMP4MuxParts` with a call to `segmentFMP4SeekAndMuxParts`, then delete the entire  `segmentFMP4MuxParts`  duplicate.

If the difference is intended, would it be smart to have that single line conditional on a flag and have a helper `func` that does the work, being called from these two stub functions . In other words, make the body `segmentFMP4SeekAndMuxParts` call `implSegmentFMP4SeekAndMuxParts` passing along all the arguments and a `true` and have  `segmentFMP4MuxParts` call `implSegmentFMP4SeekAndMuxParts` passing along all the arguments and a `false`, then in that helper `implSegmentFMP4SeekAndMuxParts` have the [_effective_ line 427](https://github.com/bluenviron/mediamtx/pull/4539/files#diff-17995db508ad7dc9d924e5aff2204e8376fcd5b8c6160eccc191fb25ed1b0fceR427) read something like this (comment syntax check, sorry)

```go
if (flag) {
   muxerDTS := int64(tfdt.BaseMediaDecodeTimeV1) - segmentStartOffsetMP4
} else {
   muxerDTS := int64(tfdt.BaseMediaDecodeTimeV1) + segmentStartOffsetMP4
}
```

IDisposable: @aler9 I'm still concerned about the potential typo commented ☝️

aler9: @IDisposable thanks for the report, I noticed the plus/minus sign issue and some other things and that's the reason why this PR is temporarily paused.

IDisposable (approved):
Much cleaner

PR #4547, opened 2025-05-22
Creation of a WebSocket server to retrieve statistics from media servers in order to have a supervision page and a preview page for all streams.

Erwan92:
Added a websocket server to retrieve media server statistics in order to create a web monitoring page (intuitive web GUI to come) and a multiview web page (to dinamically retrieve all video streams in a unique page)

### DIFF
--- a/internal/conf/conf.go
+++ b/internal/conf/conf.go
@@ -293,6 +293,11 @@ type Conf struct {
 	// SRT server
 	SRT        bool   `json:"srt"`
 	SRTAddress string `json:"srtAddress"`
+	
+	// WebSocket server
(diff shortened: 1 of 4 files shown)


JeDiE99:
maybe use zabbix/grafana ?
looks like unnesesery load for service

Erwan92:
> why websocket and not http ? maybe use zabbix/grafana ? looks like unnesesery load for service

I think that can be useful and easier for everybody to use the websocket and to directly connect to webpage i will push soon. Because I did it with http (requests limit) and grafana configuration took me many time with python script for panels. In summary, I propose this because I have already tried the rest and in practice it is the simplest. Regards

PR #4573, opened 2025-05-28
Concise path permission config

AudricY:
# Summary

This PR adds support for specifying multiple paths in a single JWT authentication permission entry, providing more flexible and concise permission management. Got help from Google Jules to do it👍.

### New Features
- **Multiple paths support**: Users can now specify multiple paths in a single permission entry using the new `paths` field
- **Backward compatibility**: Existing `path` field continues to work as before
- **Validation**: Added validation to prevent using both `path` and `paths` in the same permission entry

### API Changes
- Added `paths` field to `AuthInternalUserPermission` struct (optional array of strings)
- Changed `path` field to be a pointer (`*string`) to properly handle nil values
- Added `Validate()` method to ensure `path` and `paths` are not used together
- Updated OpenAPI specification to reflect schema changes

### Authentication Logic Updates
- Updated permission matching logic to prioritize `paths` when present
- Maintained full backward compatibility with existing `path` field

## Benefits
- **Reduced configuration complexity**: Instead of creating multiple permission entries for the same action, users can specify multiple paths in a single entry
- **Better maintainability**: Easier to manage permissions for users who need access to multiple related paths

## Testing
- Added comprehensive test suite covering various scenarios
- Updated existing tests to use the new pointer-based `path` field
- All tests pass

## Documentation
- Updated `mediamtx.yml` configuration file


aler9:
Hello, it's already possible to assign permissions to multiple paths by using regular expressions:

- action: publish
  path: ~^(path1|path2|path3)$

We generally do not want to provide two ways of doing the same thing, since such situation would introduce confusion. Another side effect is that the configuration file gets more complex than what it is, another thing that we want to avoid.

Issue #3718, opened 2024-09-02
allow to toggle recording on paths without a dedicated configuration


pikachu937:
## Which version are you using?

v1.8.4

## Which operating system are you using?

- [x] Linux amd64 standard
- [ ] Linux amd64 Docker
- [ ] Linux arm64 standard
- [ ] Linux arm64 Docker
- [ ] Linux arm7 standard
- [ ] Linux arm7 Docker
- [ ] Linux arm6 standard
- [ ] Linux arm6 Docker
- [ ] Windows amd64 standard
- [ ] Windows amd64 Docker (WSL backend)
- [ ] macOS amd64 standard
- [ ] macOS amd64 Docker
- [x] Other (LXC)

## Describe the issue

there is no way to change parameters inside the all_other path, for example, set the record to true or false for a specific stream published via ffmpeg or camera (which can publish on the server)

sending to the API to disable recording for a specific stream
127.0.0.1:9997/v3/config/paths/patch/65598042706279143845

```
{	
			"record": false
}
```
and everything would be fine if its confName parameter had the value 65598042706279143845, but since the confName parameter has the value all_others, we cannot change it for a specific stream (they differ in the **name** parameter), only for all at once, which is not correct

```
{
  "itemCount": 3,
  "pageCount": 1,
  "items": [
    {
      "name": "2222222222222222222222222222222222222222222222222222222222",
      "confName": "2222222222222222222222222222222222222222222222222222222222",
      "source": {
        "type": "rtspSource",
        "id": ""
      },
      "ready": true,
      "readyTime": "2024-09-02T10:24:46.565974242+03:00",
      "tracks": [
        "H264"
      ],
      "bytesReceived": 0,
      "bytesSent": 0,
      "readers": []
    },
    {
      "name": "655980427062791438457878",
      "confName": "all_others",
      "source": {
        "type": "rtspSession",
        "id": "76cb7ae2-62fb-4b22-a33c-db2986c0ccc2"
      },
      "ready": true,
      "readyTime": "2024-09-02T10:20:07.234365534+03:00",
      "tracks": [
        "H264"
      ],
      "bytesReceived": 21918014,
      "bytesSent": 273508,
      "readers": []
    },
    {
      "name": "mystream",
      "confName": "all_others",
      "source": {
        "type": "rtspSession",
        "id": "d2cda1ab-4b44-4a45-a760-4facff809d4f"
      },
      "ready": true,
[...]

PR #4576, opened 2025-05-28
feat(core): apply new path configs by hot reloading when exact name match is found

jererobles:
A small improvement over https://github.com/bluenviron/mediamtx/pull/4568 that closes https://github.com/bluenviron/mediamtx/issues/3718

When a new path configuration is added via the API with a name _that exactly matches_ an existing stream path then the pathConf will be set to the new path configuration. The new configuration will be applied by hot reloading if possible.

### DIFF
--- a/internal/core/path_manager.go
+++ b/internal/core/path_manager.go
@@ -54,8 +54,9 @@ type pathSetHLSServerReq struct {
 }
 
 type pathData struct {
-	path  *path
-	ready bool
+	path     *path
+	ready    bool
+	confName string
 }
 
 type pathManagerParent interface {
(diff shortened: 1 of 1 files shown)


jererobles:
This actually seems to be already fixed in https://github.com/bluenviron/mediamtx/pull/4568, merged yesterday nonetheless!

aler9:
Thanks for providing this PR, i did the following changes:

* i fixed lint errors
* i implemented the reverse process (switching from specific path configuration to generic path configuration)
* i fixed unsafe access to path configuration, since `Path.conf` belongs to the path routine and cannot be accessed by `PathManager`. I added a `pathName` member to `pathData` that belongs to `PathManager` and therefore can be safely accessed by it.

Then i merged. Thanks again.

PR #4592, opened 2025-06-01
Add a `make` target for local `mediamtx`

rjp:
Minor tweak just to make it easier to build for people who want to type `make X` instead of finding the build instructions in the middle of `README.md`.

(prompted by https://www.jwz.org/blog/2025/05/rtmp-hls-segment-corruption/#comment-259721 )


aler9:
Hello, we try to use tools that are as standard as possible, and `go generate ./... && go build` is the standard tool to build Go executables.

PR #4616, opened 2025-06-05
Addition of a websocket server which will allow to easily expose all the statistics of the multimedia servers with a view to adding a monitoring web page which will display the number of viewers, streams etc...

Erwan92:
You can enable/disable the websocket in the YAML. you just have to connect to ws://@IP:port/ws (you choose port and interval of refresh in the yaml) to have every informations about every streams. I will push a web GUI based on this websocket soon.

### DIFF
--- a/internal/conf/conf.go
+++ b/internal/conf/conf.go
@@ -294,6 +294,11 @@ type Conf struct {
 	SRT        bool   `json:"srt"`
 	SRTAddress string `json:"srtAddress"`
 
+	// WebSocket server
+	WebSocket        bool   `json:"websocket"`
+	WebSocketAddress string `json:"websocketAddress"`
+	WsInterval       int    `json:"wsInterval"`
+
 	// Record (deprecated)
 	Record                *bool         `json:"record,omitempty"`                // deprecated
 	RecordPath            *string       `json:"recordPath,omitempty"`            // deprecated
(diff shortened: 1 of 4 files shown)


JeDiE99:
For conversasion history https://github.com/bluenviron/mediamtx/pull/4547

JeDiE99:
I think more useful for highload system collect statistic in grafana + Zabbix/Prometheus
You can collect statistics localy by current http API (even by bash) and send to your moninoring system and etc
This PR requires permanent tcp connection, where WSS ?
What issue you want to solve with this patch ?

aler9:
You can run this in a separate binary that pulls data from the server's metrics interface, there's no need to change the server code.

Erwan92:
I already did it with grafana and python scripts to dynamically generate
panels but I think it would be easier for everyone to have a system that
directly exposes this information (like all software that has its own
management/supervision interface), rather than developing separate
binaries. As I initially said, this websocket was supposed to be used for a
web GUI, but in the end I’m exposing this information via SSE on a web GUI,
and the websocket is now used to interconnect multiple servers in a star
topology, as well as to control the bitrate of the devices sending the
streams (by modifying the h264 binaries to adjust the bitrate on the fly).

Regards,

Le mar. 12 août 2025 à 14:11, Alessandro Ros ***@***.***> a
écrit :

PR #4664, opened 2025-06-24
Add MPEG-TS recording support for streams with KLV metadata

zach-grant-skydio:
This commit adds support for recording MPEG-TS streams that contain KLV metadata or other unsupported elementary streams. Previously, the recorder would fail when encountering such streams with the message "no supported tracks found, skipping recording".

Key changes:
- Implement a passthrough recorder for MPEG-TS that writes raw stream data directly to disk without trying to process unsupported elementary streams
- Modify the recorder to automatically use passthrough mode when unsupported codecs are detected
- Update the MPEG-TS reader to properly extract and forward raw MPEG-TS data using Generic units with RTP packets

This allows MediaMTX to gracefully handle MPEG-TS streams with KLV metadata by preserving the original stream structure and writing it directly to disk.

### DIFF
--- a/internal/protocols/mpegts/to_stream.go
+++ b/internal/protocols/mpegts/to_stream.go
@@ -3,11 +3,13 @@ package mpegts
 
 import (
 	"errors"
+	"io"
 	"time"
 
 	"github.com/bluenviron/gortsplib/v4/pkg/description"
 	"github.com/bluenviron/gortsplib/v4/pkg/format"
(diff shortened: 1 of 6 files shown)


zach-grant-skydio:
Working through the existing tests, and also will add ones for this feature.

I wanted to go ahead and tender the feature idea though and get feedback, please let me know if this is feasible!

zach-grant-skydio:
Closing in favor of a simpler approach: https://github.com/bluenviron/mediamtx/pull/4669

PR #4668, opened 2025-06-25
h264/h265: log packet sizes if they exceed UDPMaxPayloadSize

tmatth:
This can be helpful for knowing how to tune upstream sources to avoid the need for remuxing.


tmatth:
This would be helpful for debugging scenarios like #5161

tmatth:
ping @aler9

aler9:
merged, thanks

Issue #4408, opened 2025-04-10
Issues loading WebRTC and HLS in Firefox


dnoliver:
### Which version are you using?

docker.io/bluenviron/mediamtx:1.10.0

### Which operating system are you using?

Windows amd64 Docker (WSL backend)

### Describe how to replicate the issue

1. start the server
2. Publish a ts video to the rtsp input
3. Read streaming with WebRTC and HLS

The following screenshot shows the same WebRTC stream in Chrome (left), Firefox (center), and Edge (right).
WebRTC fails to load in Firefox

![Image](https://github.com/user-attachments/assets/176993c7-aaec-4091-a5b4-9a69964dc137)

The following screenshot shows the same HLS stream in in Chrome (left), Firefox (center), and Edge (right).
HLS never finishes loading in Firefox (but works).
Programmatically, the iframe will not launch the load event
BUT (just found out) if you click in the Stop Loading button, it works fine (not sure if it trigger some event here)

![Image](https://github.com/user-attachments/assets/5c981791-35cb-4d0d-bdd4-1af370586a5e)

### Server logs

I think this is a browser side issue

aler9:
The HLS issue was fixed with #4706.

dishmaker:
For the WebRTC Firefox issue, i think adding `['H264/90000']` might help

https://github.com/bluenviron/mediamtx/blob/1bd5febcad37f9f65fac42805d2137abc1bedf5b/internal/servers/webrtc/reader.js#L374-L379

dishmaker:
I tried calling `supportsNonAdvertisedCodec(mediaType, codec, fmtp)` with 
- `mediaType = "video"`,
- `codec = "H264/90000"`,
- `fmtp` from Chrome (e.g. `level-asymmetry-allowed=1;packetization-mode=1;profile-level-id=42001f`)

without success.

dishmaker:
Okay, it's because Fedora does not have OpenH264 plugin enabled in Firefox.

<img width="691" height="401" alt="Image" src="https://github.com/user-attachments/assets/295ae090-1f4f-4a08-8693-3826ccdb438b" />

aler9:
I tried replicating the WebRTC issue but in my case everything worked flawlessly.

I used:
* Windows 11 22631
* Edge 140
* Chrome 139
* Firefox 143

And published a H264 stream with:

```
[...]

PR #4706, opened 2025-07-08
hls: fix endless loading in firefox

dishmaker:
Fixes partially:
- #4408

### DIFF
--- a/internal/servers/hls/index.html
+++ b/internal/servers/hls/index.html
@@ -195,6 +195,8 @@ const init = () => {
 	loadStream();
 };
 
+// use load instead of DOMContentLoaded, otherwise, in Firefox,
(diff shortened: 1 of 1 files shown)


aler9:
Hello, how is this supposed to fix loading? can you document this change?

dishmaker:
This just fixes the constant loading icon in Firefox tab bar.
Looks like firefox treated all network i/o before "load" event as essential to website being shown as loaded.

dishmaker:
The spinning circle icon:
Screenshot_20250712-150736~2

aler9:
you are right, in Firefox when HLS is loaded after DOMContentLoaded, the page gets stuck in the "loading" state. Thanks for noticing and fixing the issue.

PR #4707, opened 2025-07-08
webrtc: fix endless loading in firefox

dishmaker:
Fixes partially:
- #4408


aler9:
in reality, in case of WebRTC this bug does not happen, but it's good to align the behavior of WebRTC and HLS.

PR #4803, opened 2025-07-30
Add support for setting a `hlsBaseURL`

frectonz:
This PR depends on a PR in `gohlslib` which has not been merged yet.

https://github.com/bluenviron/gohlslib/pull/257

Allows users to specify a base URL that will be used to prefix all the media file URLs in the generated HLS playlist, this is useful for cases in which users may want to host their media at CDN.


aler9:
https://github.com/bluenviron/gohlslib/pull/257#issuecomment-3193784800

PR #4805, opened 2025-07-30
Update README.md

pkirk:
Added explicit ffmpeg RTMP example, because it needs GET params


aler9:
this has been added to the "Authentication" page:
https://mediamtx.org/docs/usage/authentication#rtmp

PR #4820, opened 2025-08-03
udp: listen on default multicast interface (#4565)

### LINKED ISSUE #4565: MediaMTX does not use system routing tables for multicast stream sources
mediamtx_v1.12.2_linux_amd64.tar.gz

Linux amd64 standard

### Describe issue
I welcome the possibility to choose a specific network interface in the form:
source: udp://238.0.0.1:1234?interface=eth0
It is very good feature. Thank you very much.

When use only:
source: udp://238.0.0.1:1234
MediaMTX tries to get the source on all network interfaces.

However, perhaps the current behavior when the interface at the source is not defined should be different.
For example, the vlc uses the system routing table when choosing the interface at multicast.

Test with vlc:

1. test
    $ vlc udp://@238.0.0.1:1234
    $ ip maddr
    2: eth0
    inet 238.0.0.1

2. test
    $ sudo ip route add 238.0.0.1/32 dev eth1
    $ vlc udp://@238.0.0.1:1234
    $ ip maddr
    3: eth1
    inet 238.0.0.1

I think that more tools in Linux behave like this (for example ffplay too: ffplay udp://238.0.0.1:1234).
vlc probably does something like this when choosing an interface:
$ ip route get 238.0.0.1

MediaMTX's current behaviour of trying to get video from all interfaces could then be set as follows:
source: udp://238.0.0.1:1234?interface=ALL
Although someone might be upset that it's a change in behavior.

aler9:
... instead of listening on all multicast interfaces.

Fixes #4565

### DIFF
--- a/README.md
+++ b/README.md
@@ -913,7 +913,7 @@ paths:
 
 The resulting stream is available in path `/mypath`.
 
-If the listening IP is a multicast IP, _MediaMTX_ listens for incoming multicast packets on all network interfaces. It is possible to listen on a single interface only by using the `interface` parameter:
+If the listening IP is a multicast IP, _MediaMTX_ listens for incoming multicast packets on the default interface picked by the operating system. It is possible to specify this interface manually by using the `interface` parameter:
 
 ```yml
 paths:
--- a/internal/staticsources/udp/source.go
+++ b/internal/staticsources/udp/source.go
@@ -24,6 +24,44 @@ const (
 	udpKernelReadBufferSize = 0x80000
 )
 
+func defaultInterfaceForMulticast(multicastAddr *net.UDPAddr) (*net.Interface, error) {
+	conn, err := net.Dial("udp4", multicastAddr.String())
+	if err != nil {
+		return nil, err
+	}
+	localAddr := conn.LocalAddr().(*net.UDPAddr)
+	conn.Close()
+
+	interfaces, err := net.Interfaces()
+	if err != nil {
+		return nil, err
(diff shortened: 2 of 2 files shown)


calanor:
Hi,
Since this commit, I can no longer connect to the multicast source.
I always get the error 

> 2025/10/11 18:11:43 ERR [path test] [UDP source] operation not permitted

Previously, it work fine except if I forced the interface in the multicast URL, for example:
`udp://239.190.101.30:5004?interface=eth0`

Now, since the default interface is always discovered, only the **multicast.NewSingleConn** function is called, and this function always returns the error. 

Previously, if I didn't specify the interface, **multicast.NewMultiConn** was called, which is what works correctly for me.
Now, **multicast.NewMultiConn** no longer exists in udp.go code.

This is my path:

```
  test:
    source: udp://239.190.101.30:5004
    sourceOnDemand: yes


```
This error persists until current commit.

The problem its located in **multicast.NewSingleConn** function from **gortsplib/tree/main/pkg/multicast/single_conn_lin.go**   which calls :

`syscall.SetsockoptString(sock, syscall.SOL_SOCKET, syscall.SO_BINDTODEVICE, intf.Name)`

but **SO_BINDTODEVICE** option which needs root privileges in linux to be executed.



Thanks,
Calanor

aler9:
@calanor it's useless to report bugs by commenting closed pull requests, open a bug report and provide details including name and version of operating system and all software and configuration involved.

PR #4883, opened 2025-08-17
Add disablepictureinpicture param

bp0:
Add disablepictureinpicture param to optionally hide the PiP button that floats over video

ref: https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/video#disablepictureinpicture

Note that Firefox hides the floating button, but PiP can still be chosen from the context menu, while Chrome disables the floating button and the context menu item.

I did not fully build mediamtx and test this. Only tested using web developer tools to change the running page.


aler9:
merged, thanks!

fxzzi:
how are we actually supposed to configure these options within mediamtx? I want to have the stream unmuted by default and I see that's an option here

bp0:
> how are we actually supposed to configure these options within mediamtx? I want to have the stream unmuted by default and I see that's an option here

query string; example:
`http://mediamtx.local:8888/stream/?muted=no`

PR #4888, opened 2025-08-20
fix: record cleaner doesn't delete segments under subdirs

iIIusi0n:
**Issue:**
The record cleaner doesn't delete recorded segments located in subdirectories. Currently, it only removes files from the root recordings directory but ignores nested folders.

**Expected Behavior:**
The record cleaner should recursively delete all recorded segments, including those in subdirectories.

**Current Behavior:**
Only segments in the root recordings directory are deleted. Segments in subdirectories are left behind.

**Reproduce:**
Please check the added testcase.


aler9:
Hello, this is completely wrong because it arbitrarily deletes all files in the recording folder and uses the last modified date of files to infer the recording date. Not all files in the recording folders should be considered recordings unless they match `recordPath`, and the recording date should be taken from file name and file content.

PR #4894, opened 2025-08-22
Remove redundant duplicated http middleware

TheKhanj:
Is this just redundant or am I missing something over here 😅?


aler9:
Thanks for spotting the issue.

Issue #4807, opened 2025-07-30
Unable to publish to server from PTZ camera


omkar-pai:
### Which version are you using?

v1.13.1

### Which operating system are you using?

Linux amd64 standard

### Describe how to replicate the issue

1. Use a PTZ Optics cam with the following specs.
```
Firmware version: SOC v9.1.34
Device model: F64.HI 
Webware Version: v1.5.5.HI
```
2. Send either SRT or RTMP stream to MediaMTX
3. See below logs on the server with logLevel debug. These show up continuously till i stop the output from cam. 
4. The SRT & RTMP out from the camera works fine when sending to FFmpeg set up as a listener.
5. The RTMP works fine when sending to Nginx RTMP server as well.

### Server logs

- In case of RTMP
```
2025/07/30 16:29:27 INF [RTMP] [conn 192.168.104.1:60676] opened
2025/07/30 16:29:27 INF [RTMP] [conn 192.168.104.1:60676] closed: unexpected chunk stream ID
2025/07/30 16:29:27 INF [RTMP] [conn 192.168.104.1:60678] opened
2025/07/30 16:29:27 INF [RTMP] [conn 192.168.104.1:60678] closed: unexpected chunk stream ID
2025/07/30 16:29:28 INF [RTMP] [conn 192.168.104.1:60680] opened
2025/07/30 16:29:28 INF [RTMP] [conn 192.168.104.1:60680] closed: unexpected chunk stream ID
```
- In case of SRT
```
2025/07/30 16:22:59 WAR [SRT] [conn 192.168.104.1:56622] 29 decode errors
2025/07/30 16:23:00 WAR [SRT] [conn 192.168.104.1:56622] 31 decode errors
2025/07/30 16:23:01 WAR [SRT] [conn 192.168.104.1:56622] 30 decode errors
2025/07/30 16:23:02 WAR [SRT] [conn 192.168.104.1:56622] 30 decode errors
2025/07/30 16:23:03 WAR [SRT] [conn 192.168.104.1:56622] 29 decode errors
```

### Network dump

Network dump in both cases
https://drive.google.com/drive/folders/18v6L-MSez3VZC-ZX12ugrLdagwYNtJmo?usp=sharing

aler9:
the RTMP part is fixed by #4909

PR #4909, opened 2025-08-29
feat: improve RTMP control message compatibility with non-standard cameras

hyesung913:
## Problem
MediaMTX strictly enforces RTMP specification requiring control messages to use Chunk Stream ID 2, causing connection failures with many IP cameras and PTZ cameras that use non-standard implementations.

## Solution
- Added flexible `isControlChunkStreamID()` validation function
- Support common non-standard Chunk Stream IDs (2, 3, 4, 5, 6) used by various camera manufacturers
- Maintain full backward compatibility with standard RTMP clients
- Centralized compatibility management for easy extension

## Changes
- **Core**: Added `isControlChunkStreamID()` helper function in `message.go`
- **Control Messages**: Updated all control message validation logic
- **Compatibility**: Support Chunk Stream IDs 2, 3, 4, 5, 6
- **Error Messages**: Improved error reporting with specific Chunk Stream ID values

## Benefits
- PTZ cameras (Chunk Stream ID 4) now work
- Various IP cameras (3, 5, 6) now supported  
- nginx-rtmp level compatibility achieved
- Zero impact on existing functionality
- Easy to extend for future camera types

## Background
This issue was discovered when PTZ cameras failed to connect with "unexpected chunk stream ID" errors. Investigation showed that while RTMP spec defines Chunk Stream ID 2 for control messages, many real-world camera implementations use different IDs (3, 4, 5, 6). nginx-rtmp handles this gracefully, and this PR brings MediaMTX to the same level of practical compatibility.

Fixes connection issues with cameras that send control messages on non-standard chunk stream IDs while maintaining strict RTMP compliance for standard clients.

### DIFF
--- a/internal/protocols/rtmp/message/message.go
+++ b/internal/protocols/rtmp/message/message.go
@@ -10,6 +10,26 @@ const (
 	ControlChunkStreamID = 2
 )
 
+// Common non-standard control chunk stream IDs used by various cameras
+var commonControlChunkStreamIDs = []byte{
+	2, // Standard RTMP
+	3, // Some cameras use this
+	4, // PTZ cameras and others
+	5, // Some IP cameras
+	6, // Alternative implementation
+}
+
+// isControlChunkStreamID checks if the given chunk stream ID is acceptable for control messages
+// This provides compatibility with various camera implementations that don't strictly follow RTMP spec
+func isControlChunkStreamID(chunkStreamID byte) bool {
+	for _, id := range commonControlChunkStreamIDs {
+		if chunkStreamID == id {
+			return true
+		}
+	}
+	return false
+}
+
 // Type is a message type.
 type Type byte
 
--- a/internal/protocols/rtmp/message/message.go
+++ b/internal/protocols/rtmp/message/message.go
@@ -10,26 +10,6 @@ const (
 	ControlChunkStreamID = 2
 )
 
-// Common non-standard control chunk stream IDs used by various cameras
-var commonControlChunkStreamIDs = []byte{
-	2, // Standard RTMP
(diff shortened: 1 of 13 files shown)


aler9:
Hello, can you write the manufacturer and model of the camera that require this change?

Furthermore, can you provide tests or alternatively a network dump of packets exchanged between camera and server so we can generate tests?

hyesung913:
> Hello, can you write the manufacturer and model of the camera that require this change?
> 
> Furthermore, can you provide tests or alternatively a network dump of packets exchanged between camera and server so we can generate tests?

The PTZ camera I used is:

Manufacturer: Ever Information

Model: 330UI

Link: [Product page](https://presentation.aver.com/model/md330ui?_gl=1*bpqzfi*_ga*NjQ5NDU0OTguMTc1MjgxMjI3MA..*_ga_ZCSBWSSLM4*czE3NTY3MDc0NTAkbzkkZzAkdDE3NTY3MDc0NTAkajYwJGwwJGgw&utm_source=chatgpt.com)

At the moment, I don’t have access to this camera, so I cannot capture a new packet dump myself.
However, there is an existing issue(#4807 ) describing the same problem, and that report already includes a packet dump. The attached screenshot is taken from that dump.

In the dump, the Set Chunk Size command is sent as a control message. According to the RTMP specification, it should use Chunk Stream ID 2, but this camera sends it on ID 4 instead.

It seems that some PTZ cameras do not fully comply with the standard. Both ffmpeg and nginx-rtmp are tolerant of this behavior and handle it gracefully.
image

aler9:
i removed the check entirely, added tests and merged, thanks!

PR #5013, opened 2025-09-22
Modified parseSegments to detect EOF errors

nicabreon:
Error Description

Observed Symptoms:
Log entries ERR [playback] EOF message. (2025/09/22 08:54:04 ERR [playback] EOF)
The error occurred during the execution of the onList handler in the MediaMTX playback package, specifically within the parseAndConcatenate function, which processes FMP4 (Fragmented MP4) recording segments.

Solution Proposed:
Modified parseSegments to detect EOF errors and skip the affected segments (chSeg <- nil, ch <- nil) instead of failing
Add buffered channels (ch and chSeg) with a capacity of len(segments) to prevent deadlocks when processing large numbers of segments
Used a separate chSeg channel to collect only valid parsedSegment objects, appending them to the parsed slice only if non-nil
Added a check to return an error ("no valid segments found") if all segments are skipped or fail, ensuring the API does not silently succeed with an empty result.

### DIFF
--- a/internal/playback/on_list.go
+++ b/internal/playback/on_list.go
@@ -14,6 +14,7 @@ import (
 	"github.com/bluenviron/mediamtx/internal/conf"
 	"github.com/bluenviron/mediamtx/internal/recordstore"
 	"github.com/gin-gonic/gin"
+	"io"
 )
 
 type listEntryDuration time.Duration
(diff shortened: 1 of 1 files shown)


aler9:
this doesn't solve anything, we are already skipping segments with errors, and finally this is not a clever way to use LLMs.

nicabreon:
thank you for fast response and advice,

i don't know if your statement don't solve anything is correct or not,
I make this pr because on my environment we found that mediamtx return EOF when date start not inside duration of video segment, for example start parameter is 2025-09-22T00:00:01Z and video segment filename is 2025-09-22_04-17-11-303864.mp4

for note, our video segment inside folder is about 6400 video segment with each segment 600s long,
I have tried using mediamtx version 1.11 and 1.15 it result same EOF, change folder storage, checking permission, verify manually some video, etc

List Video Segment
Screenshot 2025-09-23 at 14 02 24

image


Query Failed EOF
image

Query Success After Fix
image


Lastly, even if you don't accept my pull request, I hope you'll check back on this.
Thank you.

PR #5039, opened 2025-09-27
do not include recorder and HLS muxer in sent bytes (#4380) 

### LINKED ISSUE #4380: bytesSent counting incorrect if used as network metrics (recorder is not a network reader)
Issue Description
Currently, the recorder (which writes to local disk) is treated as a regular stream reader, causing its disk write operations (consumption of bytes from the source) to be incorrectly counted as network BytesSent. This leads to inflated network statistics that don't reflect actual network traffic.

Problem

Expected behavior: BytesSent should only count bytes sent over the network to clients (HLS, WebRTC, RTSP, etc.)
Current behavior: BytesSent includes both network traffic AND local disk writes (source data consumption) from the recorder
Impact: Network monitoring, bandwidth calculations, and statistics are inaccurate

If it is not a predicted (correct) behavior, and bytesOut is about to be used as network stats metrics, then i can prepare a PR.

aler9:
Fixes #4380

in API (/paths/list, /paths/get) and metrics (paths_bytes_sent), the amount of sent bytes was increased even in case of writes to the recorder and HLS muxer, which are not generating network traffic. This fixes the issue.


debugmenot:
Hello. Thanks for the update on my issue report. I think it will be also a good idea to count bytesOut aggregated from the client/muxers sessions/connections, maybe, grouped by protocol (rtsp/rtmp/hls/webrtc), then the stats will be more clear and informative, so we will know how much outbound traffic for each protocol is consumed, it can be very usefull when planning load balancing, determining top/popular paths, etc.

PR #5063, opened 2025-10-06
Update 17-start-on-boot.md

t-ihmels:
Updated the startup requirements for the mediamtx daemon to allow WebRTC clients to connect without continual timeouts.

### DIFF
--- a/docs/2-usage/17-start-on-boot.md
+++ b/docs/2-usage/17-start-on-boot.md
@@ -11,12 +11,6 @@ sudo mv mediamtx /usr/local/bin/
 sudo mv mediamtx.yml /usr/local/etc/
 ```
 
-Enable a _wait-online_ service:
-
-```sh
-sudo systemctl enable systemd-networkd-wait-online.service
-```
-
 Create a _systemd_ service:
 
 ```sh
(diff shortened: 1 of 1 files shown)


t-ihmels (commented):
Note that this was resolved and tested on Debian 12 / Raspberry Pi OS 64-bit Lite running on a Pi Zero 2W.

PsymoNiko (approved):
@t-ihmels Good,

also You Can Persist your added camera paths by API with updating the `mediamtx.yml` base on this [python script](https://github.com/PsymoNiko/mediamtx-dashboard/blob/main/mediamtx_updater.py), it's kind if you take a review base on the solution that I did for storing added camera pahs and configuration by `API` after system reset

aler9:
This took a while to review because i had to check whether the proposed solution is compatible with every systemd-based linux distribution. The short answer is that it is indeed, and lots of other servers and services depend on `network-online.target`, including Docker and cups.

However, enabling `systemd-networkd-wait-online.service` is required only when `systemd-networkd` is in charge of managing network, which is not the case when NetworkManager is installed. But, when NetworkManager is installed, enabling  `systemd-networkd-wait-online.service` is harmless and can be done anyway.

Therefore, this can be merged. Thanks.

PR #5125, opened 2025-10-22
docs: update

aler9:


PsymoNiko (approved):
Thanks for make it better

PsymoNiko (approved):
Good Point, Thanks For Fixing

PsymoNiko:
@aler9 Maybe Its better to use plain UTF-8 text to pass the tests
BWT, thanks for make its environment better day by day

PR #5136, opened 2025-10-27
Doc: Clarify network settings for RTSP on macOS and Windows

saint-james-fr:
Added note about port binding for macOS and Windows.


aler9:
Thanks for specifying that the `--network=host` flag is not compatible with macOS too. Apart from this, the new comment is a repetition of the paragraph below. In #5125 i've moved the command line without `--network=host` above the one with `--network=host`, and added the macOS note.

PR #5150, opened 2025-11-02
feat(api): add support for multiple allowed origins with wildcards

### LINKED ISSUE #5091: support multiple HTTP origins
### Describe the feature

Discussed in #4712 

reference: https://stackoverflow.com/questions/1653308/access-control-allow-origin-multiple-origin-domains

In order to support multiple HTTP origins, there's need for a dedicated mechanism that
* reads the Origin header
* checks it against a whitelist
* if present, puts its value inside Access-Control-Allow-Origin

secit:
Introduce a new `allowOrigins` configuration field (slice of strings) to replace the single-string `allowOrigin`.
Add logic to validate origins against allowed patterns, supporting:
- Exact matches
- Wildcard domain matching (e.g., http://*.example.com)

Update configuration handling in core and test components to support the new format.
Maintain backwards compatibility with the old `allowOrigin` field.
Update `openapi.yaml` to reflect the new configuration field.

Fixes #5091

### DIFF
--- a/api/openapi.yaml
+++ b/api/openapi.yaml
@@ -144,8 +144,10 @@ components:
           type: string
         metricsServerCert:
           type: string
-        metricsAllowOrigin:
-          type: string
+        metricsAllowOrigins:
+          type: array
+          items:
+            type: string
         metricsTrustedProxies:
           type: array
           items:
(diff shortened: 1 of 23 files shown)


aler9:
i extended the feature to every HTTP-based service and merged. Thank you very much!

PR #5156, opened 2025-11-04
Update 1-configuration-file.md

PsymoNiko:
There is no path of the `1-configuration-file` also I think it's better to set a link to avoid mistake during the redirect to the exact folder or file

Screenshot_20251104_221844_Chrome


PsymoNiko (commented):
I checked on myself 
Wish It'll be acceptable

PsymoNiko (commented):
To redirect to the exact file path

aler9:
the documentation is meant to be rendered on https://mediamtx.org, and these changes would cause a huge amount of issues.

PsymoNiko:
> the documentation is meant to be rendered on https://mediamtx.org, and these changes would cause a huge amount of issues.

Oh sorry,but thanks for taking a look on my PR 🫶
Next, I'll to do it better🤝

PR #5166, opened 2025-11-07
test: fix and assert that all tracks are recorded

boenshao:
In the `TestRecorder` test, the H.265 track has not been recorded since commit 253df883.

image

The issue appears to be with the `test.FormatH265` sample. Hard-coding the sample as in the `TestFromStream` test resolves the problem.


aler9:
i refactored and merged, thanks

PR #5196, opened 2025-11-14
api: add keepalive endpoints to maintain stream connections

thair116:
Add keepalive API endpoints that allow streams to be kept active without real viewers. Keepalives act as synthetic readers that trigger on-demand publishers and prevent streams from closing when all real viewers disconnect. This allows streams to be toggled into a "locked on" state, even if all readers drop off, while running in an on-demand-only mode by default. 

Before this PR, as far as I can tell this can only be accomplished for `sourceOnDemand` streams by toggling that via API, but cannot be accomplished for `runOnDemand` published streams without using a dummy reader that will a) consume resources and b) have to be managed out-of-band to mediamtx. Whereas this solution works for any stream, regardless of source/publish

Features:
- POST /v3/paths/keepalive/add/{name} - create keepalive for path
- DELETE /v3/paths/keepalive/remove/{id} - remove keepalive by ID
- GET /v3/paths/keepalive/list - list all active keepalives
- GET /v3/paths/keepalive/get/{id} - get keepalive details

Implementation details:
- Keepalives implement defs.Reader interface
- Full authentication and authorization support
- Ownership tracking - only creator can remove keepalive
- Automatic cleanup when paths are removed
- UUID-based identification
- Asynchronous stream initialization to prevent deadlocks

### DIFF
--- a/api/openapi.yaml
+++ b/api/openapi.yaml
@@ -668,6 +668,7 @@ components:
           - rtspsSession
           - srtConn
           - webRTCSession
+          - keepalive
         id:
           type: string
 
@@ -727,6 +728,36 @@ components:
       properties:
         start:
           type: string
+
+    Keepalive:
+      type: object
+      properties:
+        id:
+          type: string
+          format: uuid
+        created:
+          type: string
+          format: date-time
+        path:
+          type: string
+        creatorUser:
+          type: string
+        creatorIP:
+          type: string
+
+    KeepaliveList:
+      type: object
+      properties:
+        pageCount:
+          type: integer
+          format: int64
+        itemCount:
+          type: integer
+          format: int64
+        items:
+          type: array
+          items:
+            $ref: '#/components/schemas/Keepalive'
 
     RTMPConn:
       type: object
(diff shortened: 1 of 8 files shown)


JeDiE99:
Maybe more usefull add boolean param for path that always ignites path instead of entity+api "keepalive" ?

JeDiE99:
if True - add fake reader, false - not fake reader

thair116:
If I understand you correctly, you're suggesting to add an optional boolean param to path (e.g. keepalive=true) such that the path would stay hot, even if/after that requester disconnects? 

That would seem to handle the "lock this stream on" usecase, but what about the corresponding "close out the stream" side of things? It would be a bit awkward to have to call the path endpoint again for that

Or perhaps you're saying the original requester (with keepalive=true) simply must stay connected to keep the stream flowing? (with no bytes towards the keepalive requester, thanks to fake reader)

I'd be concerned about the unnecessary brittleness created by that connection needing to be kept alive

Please let me know further what you're thinking

aler9:
Hello, the feature is clear but i can't understand its usefulness - basically you have a path with `sourceOnDemand` set to true, which makes the server connect to the source only when there's at least a reader, but at the same time you want the connection to remain open indefinitely. Isn't it better to use `sourceOnDemand: false` from the beginning?

thair116:
The problem with `sourceOnDemand:false` is that for my usecase I have a very high number of cameras that would swamp the network if they were always running, so I need to dynamically turn on and off.

The usual flow basically looks like this
1. All cameras off (baseline - conserve network)
2. Something drives interest towards a particular camera stream. A user can start viewing that stream, and then while viewing, that camera stream will be on (thanks to `runOnDemand`) 
3. User then wants to shut down their computer but keep that stream running (say, because they need sleep)
4. Now this is where I post a `keepalive` for that stream to ensure it stays on, until the user is fully done with it

aler9:
Instead of hacking the server, my advice is to use a dummy reader (you can build it with gortsplib or just use ffmpeg) that is triggered with `runOnReady` and periodically connects to the server when you need to keep a certain stream open, achieving the same goal as this PR but without the effort of developing and maintaining 4 new endpoints.

PR #5210, opened 2025-11-18
Fix OpenAPI/Swagger double number format

FerroO2000:
Change the format for properties of type `number` from `float64` to `double`.
As stated by the OpenAPI/Swagger v3 [spec](https://swagger.io/docs/specification/v3_0/data-models/data-types/#numbers) there isn't a `float64` format for numbers.
This allows code generation tools like [openapi-go](https://github.com/swaggest/openapi-go) to generate the proper types/stubs.

### DIFF
--- a/api/openapi.yaml
+++ b/api/openapi.yaml
@@ -432,16 +432,16 @@ components:
           type: boolean
         rpiCameraBrightness:
           type: number
-          format: float64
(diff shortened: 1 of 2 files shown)


aler9:
i checked the specification and merged, thanks!

PR #5232, opened 2025-11-27
Add per path metadata and the ability to search paths list

eran2903:
This update adds two improvements to the MediaMTX paths API.

1. Per-path metadata
Each path can now include optional metadata fields in its configuration, such as expected bitrate or camera origin. These values are returned in the /v3/paths/list response so that monitoring tools or clients can understand the expected characteristics of each stream.

2. Path search support
The /v3/paths/list endpoint now supports a "search" query parameter. It filters the results and returns only the paths whose name starts with the given string. This is useful for grouping paths by camera or any naming prefix.

Example:
curl 'http://127.0.0.1:9997/v3/paths/list?search=cam2'
returns only cam2/main and cam2/sub.

Calling /v3/paths/list without the search parameter returns all paths.

### DIFF
--- a/internal/api/api.go
+++ b/internal/api/api.go
@@ -560,6 +560,18 @@ func (a *API) onPathsList(ctx *gin.Context) {
 		return
 	}
 
+	// Filter by search parameter if provided
+	search := ctx.Query("search")
+	if search != "" {
+		filteredItems := make([]*defs.APIPath, 0, len(data.Items))
+		for _, item := range data.Items {
+			if strings.Contains(strings.ToLower(item.Name), strings.ToLower(search)) {
+				filteredItems = append(filteredItems, item)
+			}
+		}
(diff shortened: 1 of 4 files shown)


Snafu:
This could also be used to solve issue #5406, instead of the pull request #5405 I made, if the metadata were to be made available as an environment variable for the runOnInit, runOnReady, runOnNotReady, runOnDemand and runOnUnDemand hook commands.

aler9:
https://github.com/bluenviron/mediamtx/pull/5405#issuecomment-4026222525

aler9:
@eran2903 the search feature is OK and it would have been merged immediately if you'd submitted that in a dedicated PR.

aler9:
this cannot be merged for reasons explained in https://github.com/bluenviron/mediamtx/pull/5405#issuecomment-4026222525

PR #5245, opened 2025-12-04
chore: update .gitignore to ignore build artifacts and editor files

HoangDinhBui:


aler9:
Sorry but we're not interested in this. Also, MQTT-based signaling is non standard and requires an additional component, while WHEP-based signaling is universally recognized and is supported natively.

Issue #4999, opened 2025-09-20
unix+rtp publish h264 stream，decode error；using same stream，udp+rtp is ok


unv274:
### Which version are you using?

V1.1.4.0

### Which operating system are you using?

Linux arm64 standard

### Describe how to replicate the issue

1. arm64，not docker
2.may be unix domain socket using stream mode, mediamtx receive more than one packets once,so parse rtp error.

### Server logs

2025/09/20 15:47:15 INF MediaMTX v1.14.0
2025/09/20 15:47:15 INF configuration loaded from /config/mediamtx.yml
2025/09/20 15:47:15 DEB [path channel_0_main] created
2025/09/20 15:47:15 DEB [path example] created
2025/09/20 15:47:15 DEB path manager created
2025/09/20 15:47:15 INF [RTSP] listener opened on :8554 (TCP)
2025/09/20 15:47:15 INF [path channel_0_main] [RTP source] started
2025/09/20 15:47:15 INF [WebRTC] listener opened on :8889 (HTTP), :8189 (ICE/TCP)
2025/09/20 15:47:15 DEB [path channel_0_main] [RTP source] connecting
2025/09/20 15:47:15 INF [path channel_0_main] [RTP source] ready: 1 track (H264)
2025/09/20 15:47:16 INF [path channel_0_main] RTP packets are too big, remuxing them into smaller ones
2025/09/20 15:47:16 WAR [path channel_0_main] [RTP source] 163 decode errors
2025/09/20 15:47:16 WAR [path channel_0_main] 5 processing errors
2025/09/20 15:47:17 WAR [path channel_0_main] [RTP source] 90 decode errors
2025/09/20 15:47:17 WAR [path channel_0_main] 137 processing errors
2025/09/20 15:47:18 WAR [path channel_0_main] [RTP source] 71 decode errors
2025/09/20 15:47:18 WAR [path channel_0_main] 1 processing error
2025/09/20 15:47:19 WAR [path channel_0_main] [RTP source] 62 decode errors
2025/09/20 15:47:19 WAR [path channel_0_main] 103 processing errors
2025/09/20 15:47:20 WAR [path channel_0_main] [RTP source] 53 decode errors
2025/09/20 15:47:20 WAR [path channel_0_main] 5 processing errors
2025/09/20 15:47:21 WAR [path channel_0_main] [RTP source] 39 decode errors
2025/09/20 15:47:21 WAR [path channel_0_main] 136 processing errors
[...]

PR #5318, opened 2026-01-05
Add support unix socket types SOCK_SEQPACKET and SOCK_DGRAM

aperture-it:
As mentioned in issue #4999, the socket with type SOCK_STREAM causes packet merging and decoding errors. Adding the SOCK_DGRAM and SOCK_SEQPACKET types ensures record boundaries, allowing packets to be read correctly one at a time.

### DIFF
--- a/docs/2-usage/02-publish.md
+++ b/docs/2-usage/02-publish.md
@@ -237,6 +237,7 @@ paths:
   mypath:
     source: unix+mpegts:///tmp/socket.sock
 ```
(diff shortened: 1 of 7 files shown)


aler9:
Hello, the analysis is correct, the solution is not. The bug in `unix+mpegts` and `unix+rtp` is left untouched, while four source types are added with unspecified behavior and no documentation. Just edit `unix+mpegts` and `unix+rtp` to use unixgram. Thanks.

aperture-it:
I'm not claiming there are bugs in `unix+mpegts` and `unix+rtp`. It's possible that some publishers work fine with them, and if they are changed, these publishers will no longer be compatible with the socket type. Four new sources add the ability to publish media streams with other socket types, which in some cases fixes packet merging at the transport layer. I'll add documentation.

aler9:
replaced by https://github.com/bluenviron/mediamtx/pull/5351

PR #5335, opened 2026-01-10
 add "always available" streams (#2214)

### LINKED ISSUE #2214: Allow a publisher to reconnect without disconnecting readers

aler9:
Fixes #2214 

This patch allows streams to be always available to readers, generating continuous streams regardless of whether the publisher is online or not. This works by seamlessly merging together "online" streams with an "offline" video, without decoding or re-encoding anything (as always), acting on already-encoded packets. When a publisher is offline, a blank video is returned:

offline

The feature can be enabled by setting the `alwaysAvailable` property on specific paths. Track list is also mandatory:

```yml

  # Enable always-available mode, in which a video is played on repeat when the stream is not available.
  alwaysAvailable: false
  # Path to the MP4 file that is played on repeat. If not provided, a default video will be used instead.
  alwaysAvailableFile: './h264.mp4'
  # If alwaysAvailableFile is not provided, these are the tracks of the default video.
  alwaysAvailableTracks:
    # Available values are: AV1, VP9, H265, H264, Opus, MPEG4Audio, G711, LPCM
    - codec: H264
      # in case of MPEG4Audio, G711, LPCM, sampleRate and ChannelCount must be provided too.
      #  sampleRate: 48000
      #  channelCount: 2
      #  in case of G711, muLaw must be provided too.
      #  muLaw: false

```

TODO:

* [x] support multitrack streams
* [x] support audio tracks
* [x] allow to specify an alternative offline video
* [x] add documentation
* [x] add tests
* [x] deprecate "fallback"


roger-:
Awesome! Can this be extended to support an arbitrary "offline" image? Having the ability to dynamically set the image would be great too.

aler9:
@roger- absolutely, i'll add some settings like

```yml
alwaysAvailableOfflineVideo: myvideo.mp4
```

roger-:
Would it catch a change in myvideo.mp4 and reload? 

I'd like to be able to repeat the last frame until input stream resumed, and I can hack together something if myvideo.mp4 isn't cached. FYI my use case is turning an intermittent (battery) camera stream into a continuous one.

jmduerk:
When using this feature, will the paths/list API always show active for a stream with this configured, or will API return true online/offline status?

aler9:
> When using this feature, will the paths/list API always show active for a stream with this configured, or will API return true online/offline status?

i've added two different fields for this:

* the `available` field tells whether a stream is available regardless of the fact that the publisher is online or not.
* the `online` field tells whether the stream publisher is online or not.

There are also `availableTime` and `onlineTime`. Regarding the existing `ready` and `readyTime` fields, they have been kept for compatibility reasons but they are equal to `available` and `availableTime`.

aler9:
> Would it catch a change in myvideo.mp4 and reload?

yes, this is implemented automatically thanks to the already-existing hot reloading feature.

> I'd like to be able to repeat the last frame until input stream resumed, and I can hack together something if myvideo.mp4 isn't cached. FYI my use case is turning an intermittent (battery) camera stream into a continuous one.

Unfortunately this is not possible, since the "repeated" frame, which is the first frame of the offline video, must be a "random access" frame, and only a small percentages of frames are, since videos are encoded more or less in this way:

I P P P P P P P I P P P P P

(where I is an I-frame/random access, P is a progressive one)

roger-:
> Unfortunately this is not possible, since the "repeated" frame, which is the first frame of the offline video, must be a "random access" frame, and only a small percentages of frames are, since videos are encoded more or less in this way:
> 
> I P P P P P P P I P P P P P
> 
> (where I is an I-frame/random access, P is a progressive one)

Thanks. My idea is this (I do it already with ffmpeg/concat in a very hack way):

1. RTSP publish a clip to mediamtx via ffmpeg
2. Extract the last frame, duplicate and turn it into a short clip(~1s) via ffmpeg and replace current `alwaysAvailableOfflineVideo` 
3. Wait for next clip and repeat

Would that work?

If mediamtx could just repeat the last I-frame then that would simplify things, but it might be too specific a use case.

JustJoostNL:
> @roger- absolutely, i'll add some settings like
> 
> ```yaml
> alwaysAvailableOfflineVideo: myvideo.mp4
> ```

That would be awesome! Would it support other HLS/DASH streams as offline video too? Or a script that it triggers?

nrbnlulu:
does SPS/VPS/VPS params stay the same as the live stream somehow? if not, it require a keyframe to be sent on every disconnect am I right?

aler9:
> That would be awesome! Would it support other HLS/DASH streams as offline video too? Or a script that it triggers?

The issue with this system is that the offline video must be immediately available, and its first frame must be a key frame, in order to immediately replace the old stream - these two features are not guaranteed to be present on an external stream. Maybe in the future we can think about how to integrate external streams.

> does SPS/VPS/VPS params stay the same as the live stream somehow? if not, it require a keyframe to be sent on every disconnect am I right?

We already support H265 streams in which SPS/PPS/VPS params change over time. This happens for instance on cameras when you change the video settings. In case of the offline video, SPS/PPS/VPS will mostly change since they are the ones of the offline video. When they change, they are sent back to readers. The first frame in the offline video is always a keyframe, that resets everything before it.

JustJoostNL:
> The issue with this system is that the offline video must be immediately available, and its first frame must be a key frame, in order to immediately replace the old stream - these two features are not guaranteed to be present on an external stream. Maybe in the future we can think about how to integrate external streams.

Ah I understand that. But if this replaces the fallback feature, which did support HLS/DASH/script trigger, we'd loose that feature, right? Or maybe it is possible to allow specifying another path as always available fallback?

aler9:
>  ah I understand that. But if this replaces the fallback feature, which did support HLS/DASH/script trigger, we'd loose that feature, right? Or maybe it is possible to allow specifying another path as always available fallback?

The fallback feature worked at the beginning of a stream, in which a player expects some delay before receiving key frames, therefore any kind of alternative stream can be used, with the major drawback of not being able to go back to the original stream.

The Always Available feature works at any point of the stream, and pauses are not allowed, since they would trigger a timeout in players. Therefore, the alternative stream must have additional requirements (being immediately available and starting with a key frame).

Therefore yes, there are constraints on the alternative stream. Said that, most feedbacks regarding the fallback feature mentioned the inability to go back to the original stream, and that is something that we're finally fixing here. Leaving the fallback feature available would cause confusion, since there would be two features with the same scope.

Using any kind of fallback stream is still possible by using some kind of hook-based logic in which when the main publisher is not offline, the fallback stream is published in its place.

jmduerk:
Can you tell me if the new release would have also merged the upstream commit https://github.com/datarhei/gosrt/commit/03ab903b5c6a1a32870f2d368f2439f368fef5f5 where gosrt default PEERIDLETIMEO was changed?  It looks like MediaMTX uses a fork that you authored.

fragtion:
> @roger- absolutely, i'll add some settings like
> 
> ```yaml
> alwaysAvailableOfflineVideo: myvideo.mp4
> ```

`2026/02/02 23:52:06 INF [WebRTC] [session 7acb84fd] closed: want to publish [Opus H264], but stream expects [H264]`

How to deal with this? Looks like this feature isn't ready for WHIP streams that require Opus audio

PR #5342, opened 2026-01-13
Add truststore support for auth requests and check complete certificate chain with fingerprint

LukasLeppich:
This PR adds a new config property to provide a trust store in PEM format for auth HTTP calls.

New config property:
```yaml
# As an alternative to the fingerprint, you can also provide the path to a trust store in PEM format.
# The certificates will be merged with the system trust store.
# The trust store is only used if the authJWTJWKSFingerprint is not set.
authJWTTrustStorePem:
```

My first attempt to solve my certificate problem was to check the complete certificate chain for the provided fingerprint. Even though I don't use this feature, it could be useful for others, so I added the code as well.

I tried to identify all the relevant places for the configuration change, but as I am not familiar with the project, it is possible that I have overlooked something.
If anything is missing, please let me know and I will add it to the pull request.


aler9:
Hello, we already support at least two other methods to validate certificates:
* certificates authorities installed in the operating system
* fingerprint in configuration

Adding a third one is redundant. If you want to use dynamic certificates or a list of certificates, sign them with a certificate authority and add the public key of the certificate authority to the system directory (`/usr/share/ca-certificates` on Linux).

LukasLeppich:
Hello, thank you for looking into this PR.

I use the mediamtx server in an environment where I don’t have access to the system trust store and the certificate of the authentication server is short lived. Therefore, I must trust the root certificate, which I cannot check using the fingerprint.

That’s the reason why I added the trust store support.
But I also understand that you don’t want to add technical depth to your project.

Thank you for your work on this project!

PR #5363, opened 2026-01-21
Enable Restricted RTP/RTCP Source UDP ClientPort Ranges to facilitate MediaMTX Source UDP reception inside K8s deployments 

bjornss-scaleaq:
This is a suggested extension of MediaMTX Config / Env to allow reception of RTP/RTCP data over UDP, for containerized deployments used INSIDE K8s clusters, with K8s services (NodePorts/Loadbalancers ) to control UDP traffic. 

See the accompanying PR 
[(https://github.com/bluenviron/gortsplib/pull/985)]
for technical detail

The PR introduces a new Global Conf parameter / (ENV Variable) for RTSP sources:

```
# Source UDP RTP Client Port range
# This can be used to control RTSP Source UDP client port ranges for use e.g inside Kubernetes clusters
# Starting and ending ports should be EVEN numbers
# It defaults to the range [ 10000, 65534 ]
sourceUDPClientPortRange: [ 10000,65534 ]
```


aler9:
The issue with this patch is that the new parameter is treated as a global variable, creating a new parameter group, while it should be a path variable, grouped with existing RTSP parameters. This has been reimplemented with #5397

PR #5370, opened 2026-01-22
documentation changes for ffmpeg

RogerHardiman:
There was a mistake where the address and port in the ffmpeg example did not match the config file settings.
Also added example for sending KLV data

### DIFF
--- a/docs/2-usage/02-publish.md
+++ b/docs/2-usage/02-publish.md
@@ -505,7 +505,7 @@ WARNING: in case of FFmpeg 8.0, a video track and an audio track must both be pr
 
 ### GStreamer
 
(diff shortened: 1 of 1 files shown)


aler9:
Hello, while the first change fixes a typo and can be merged, the second one is more of a FFmpeg issue and is not related to the specific section (FFmpeg and MPEG-TS over UDP) - can you remove it?

RogerHardiman:
@aler9 .   Thank you for the feedback.
I've removed the 2nd commit (KLV)
I have found one other typo, so I've added that in. So there are still 2 commits, but the 2nd one is new.
Many thanks, Roger.

aler9:
merged, thanks

Issue #1420, opened 2023-01-26
LL-HLS without cert requirement / https, for use behind reverse proxy or on edge servers


renzo-s:
## Describe the feature

Is it possible to not let the service check for certificates when serving an LL-HLS stream? It's pretty common to let edge servers or reverse proxy's manage this. Also scaling in such an architecture is far more difficult when the app needs the certs at init.

Thanks!

aler9:
HTTPS with LL-HLS is currently mandatory since it's mandatory on Apply devices, but this may be a good enhancement.

aler9:
fixed by #1530

aler9:
added in v0.21.6

PR #5379, opened 2026-01-25
Update hlsEncryption documentation on Low-Latency HLS requirements

fdelu:
Updates the comments on the configuration file for the hlsEncryption property to reflect the changes in #1420 

I initially thought Low-Latency HLS wasn't supported without encryption until I found that pull request


aler9:
merged, thanks

PR #5412, opened 2026-02-02
🐛 fix: Add field name to slice nil errors

Zaphkiel-Ivanovna:
## What changed
- Error message for null slices now includes the field name

## Why
- `cannot set slice to nil` was unclear - users couldn't identify which field caused the error

## How it works
- Field path is tracked during JSON unmarshaling and included in error messages

## Example

Before:
```
ERR: cannot set slice to nil
```

After:
```
ERR: cannot set slice to nil: field 'webrtcICEServers2'
```

## Testing
- Unit tests updated and passing

### DIFF
--- a/internal/conf/jsonwrapper/unmarshal.go
+++ b/internal/conf/jsonwrapper/unmarshal.go
@@ -15,10 +15,13 @@ import (
 // - prevents using existing elements of slices, fixing https://github.com/golang/go/issues/21092
 // - prevents setting slices to nil
 
-func process(v reflect.Value, raw any) error {
+func process(v reflect.Value, raw any, path string) error {
 	switch v.Kind() {
 	case reflect.Slice:
 		if raw == nil {
+			if path != "" {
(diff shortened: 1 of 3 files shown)


aler9:
Hello, the change that includes the field name in error messages can be merged, while we don't want to add and maintain an additional command line flag. Please edit the code to include only the first change. Furthermore, TestUnmarshalSetNestedSliceToNil can be merged into TestUnmarshalSetSliceToNil. Thanks.

Zaphkiel-Ivanovna:
> Hello, the change that includes the field name in error messages can be merged, while we don't want to add and maintain an additional command line flag. Please edit the code to include only the first change. Furthermore, TestUnmarshalSetNestedSliceToNil can be merged into TestUnmarshalSetSliceToNil. Thanks.

Hello @aler9 

I've applied the requested changes.

Thanks for the review!

aler9:
i had to make some changes but unfortunately editing this patch is not allowed. Moved to https://github.com/bluenviron/mediamtx/pull/5433

Issue #5428, opened 2026-02-07
Add srtPeerIdleTimeout configuration option


fyzali:
### Describe the feature

Hi,

The SRT listener configuration is currently hardcoded and cannot be changed via the configuration file. Specifically, the PeerIdleTimeout parameter from the underlying gosrt library is not exposed in mediamtx.yml. Currently, it uses the hardcoded default of 5 seconds.

It would be great if this were configurable, especially for setups dealing with unstable networks where connections drop too aggressively. I am currently using MediaMTX as an SRT relay, and disconnections of just a few seconds are breaking the stream. This is contrary to expected SRT behavior, which can have latency of greater than 5 seconds.

For reference, gosrt defines the parameter here:

https://github.com/datarhei/gosrt/blob/main/config.go#L205

This parameter is applied when the listener is initialized in the SRT server.

Is there any chance this configuration option could be added?

Thanks!

fyzali:
Additionally, would it be possible to implement peeridletimeo similarly to the latency parameter, so that the higher value between the client and server is respected on a per-stream basis?

jmduerk:
> ### Describe the feature
> 
> Hi,
> 
> The SRT listener configuration is currently hardcoded and cannot be changed via the configuration file. Specifically, the PeerIdleTimeout parameter from the underlying gosrt library is not exposed in mediamtx.yml. Currently, it uses the hardcoded default of 5 seconds.
> 
> It would be great if this were configurable, especially for setups dealing with unstable networks where connections drop too aggressively. I am currently using MediaMTX as an SRT relay, and disconnections of just a few seconds are breaking the stream. This is contrary to expected SRT behavior, which can have latency of greater than 5 seconds.
> 
> For reference, gosrt defines the parameter here:
> 
> https://github.com/datarhei/gosrt/blob/main/config.go#L205
> 
> This parameter is applied when the listener is initialized in the SRT server.
> 
> Is there any chance this configuration option could be added?
> 
> Thanks!

MediaMTX actually uses an earlier version of gosrt where the peeridletimeo was two seconds.  I opened an issue on the upstream gosrt repo to have it changed to Haivision's default of 5s, but it hasn't made it downstream.  This is a highly needed feature in MediaMTX!

aler9:
fixed by https://github.com/bluenviron/mediamtx/pull/5489

PR #5489, opened 2026-02-19
SRT PeerIdleTimeout follows MTX readTimeout

jmduerk:
This PR would address issue #5428 by assigning the default SRT PeerIdleTimeout to mimic MediaMTX's configured readTimeout alongside ConnectionTimeout.  The upstream gosrt library uses 2s, which severely limits SRT's effectiveness in the lossy environments it was designed to work in.  Mimicking readTimeout is an elegant way to address the issue since PeerIdleTimeout is virtually a reader timeout with regards to SRT.


aler9:
merged, thanks!

PR #5546, opened 2026-03-02
Add playback OpenAPI definition

### LINKED ISSUE #5545: OpenAPI definition is missing duration and url for RecordingSegment

9f94bd1

Other (please describe)

### Describe the issue

I'm generating an OpenAPI client for mediamtx but currently RecordingSegment is missing duration and URL for the recording API

Though perhaps I need to use the playback API to get this information? Kind of confusing. It would be useful to also have an openapi description of the playback API

bouk:
Makes it easy to generate a client for it

Fixes #5545


aler9:
i renamed the file, aligned the OpenAPI definition with the latest code and merged. Thanks for suggesting this change and for implementing it.

By the way, in the future it might be better to merge the Control API and Playback API into a single one, right now it's unfeasible because there is no micro-profiling system that would allow isolated access to each feature.

PR #5552, opened 2026-03-05
fix: avoid buffering HTTP response body in loggerWriter

JulienCossec:
loggerWriter was shadow-copying every response byte into a bytes.Buffer to report the body size, causing the entire response to be accumulated in memory for the lifetime of each request. Replace the buffer with a plain int counter since dump() only ever reported the byte count anyway.


aler9:
merged, thanks!

Issue #4273, opened 2025-02-22
Support forwarding streams to other servers natively


aler9:
### Describe the feature

Discussed in #2879 

This feature is already available by using FFmpeg or GStreamer inside `runOnReady`, as described in the [README](https://github.com/bluenviron/mediamtx?tab=readme-ov-file#forward-streams-to-other-servers).

Supporting it natively would improve efficiency and decrease dependency on external software.

Sample configuration:

```yml
paths:
  mypath:
    source: rtsp://my-camera
    forward:
    - dest: whip://other-webrtc-server/other-whip-endpoint
    - dest: rtsp://other-rtsp-server/path
```

neilyoung:
It would be nice to have a means to make MediaMTX kind of "WHIP egress" capable. That would allow a lot of interesting use cases. At the moment this can be covered by `runOnReady` and `whip-client`, https://github.com/meetecho/simple-whip-client, which works nicely, but it requires a recent installation of GStreamer which includes `webrtcbin` and `nice`- components you will mostly only get by source installations or from docker.

gainggs:
Is this feature still planned, or are we going to have to rely on FFmpeg or GStreamer?

aler9:
implemented with #5558

neilyoung:
Thanks, but unfortunately no WHIP.... Nevermind.

aler9:
@neilyoung WHIP can be added easily, it was not done in the main patch because there were already too many lines of code and too many things to review. I'm willing to review and accept patches that add all the side features that are now easy to implement, from additional forward protocols to additional forward parameters.

neilyoung:
@aler9 understandable. No pressure

PR #5558, opened 2026-03-09
Add native stream forwarding support

aniolpages:
This PR adds push targets to paths.

A push target allows MediaMTX to republish a path to an external destination, with runtime status exposed through the API. Push targets can be defined statically in path configuration and can also be managed dynamically through new API endpoints.

**Supported target protocols:**

- RTMP
- RTMPS
- RTSP
- RTSPS
- SRT

**Path-level push target support**
Added a new pushTargets path option.

This allows configuring one or more outbound destinations directly in a path configuration, for example:
```yaml
paths:
  mypath:
    pushTargets:
      - url: rtmp://example.com/live/stream
      - url: rtmps://example.com/live/stream
```

At startup, configured targets are created automatically and tied to the path.

**Runtime API for push targets**
Added new API endpoints to list, inspect, add, and remove push targets for a path:

GET /v3/paths/pushtargets/list/pathName

GET /v3/paths/pushtargets/get/pathName/targetId

POST /v3/paths/pushtargets/add/pathName

```json
{
  "url":"rtsp://example.com/stream"
}
```

DELETE /v3/paths/pushtargets/remove/pathname/targetId


aler9:
This, if correctly implemented, would fix #4273

krsna1729:
Will be helpful @aler9 . 90% is rebroadcast. will save a lot of cpu, memory and orchestration. I even started looking into dynamic addition deletion ability of gstreamer to save cpu and memory. Kept switching to SRS, which handles this natively, as the last resort for now (Update: Doesn't really support via API). SRT bonding feature and no "too many reordered frames"  issues (SRT->RTMP) are bonuses as well.

AlexFreik:
Hello! Any update on this PR? It is exactly the feature we are missing the most in MediaMTX. After building MefiaMTX with this patch the CPU and RAM usage of our server dropped several folds compared to custom FFMPEG jobs to dynamically forward streams. When there is 300+ forward streams it makes a huge difference

aniolpages:
Since most major streaming platforms use port 443 for RTMPS, I've changed the default port (when none is specified) to 443, even though this differs from MediaMTX's default RTMPS server port.

What do you think about this, @aler9?

krsna1729:
With help from agents this is the upgraded version of the patch we are testing - 
https://github.com/live-miracles/restream/blob/master/patches/mediamtx/mediamtx_push_targets_v1.17.1.patch

aler9:
Hello, i think we can move things forward (literally in this case :D). This feature can be merged but requires a couple of architectural and implementation changes:

* rename `pushTarget` into `forward`, and `url` into dest, in order to produce configurations like these:

   ```yml
   forward:
   - dest: rtmp://
   - dest: rtsp://
   ```

   in this way we won't introduce an additional term for defining the same action, since we're already mentioning "forward" in the documentation, and there's no sense in introducing "push targets" since they are the same thing.

* rename the `push` folder into `forward`

* rename `target` into `dest` - in this way it mirrors the existing `source` subsystem, which provides the opposite feature (pulling streams).

* `Target / Dest` must be specialized - right now it's an object that contains both code shared between all targets, and protocol-specific code. Rename it to `DestHandler`, which is a class that will contain the shared code, then define `Dest` as an interface, and define a target for each protocol, obtaining this hierarchy:

   ```
   Path
   * ForwardManager (internal/forward/manager.go)
      * DestHandler (internal/forward/dest_handler.go) (struct) (one for each destination)
         * Dest (internal/forward/dest.go) (interface, implemented by)
            * rtmp.Dest (internal/forward/rtmp/dest.go)
            * rtsp.Dest (internal/forward/rtsp/dest.go)
            * srt.Dest (internal/forward/srt/dest.go)
   ```

Feel free to suggest changes or improvements. Thanks for providing this patch.

aniolpages:
Hi, 
I’ve made the requested changes. `pushTarget` is now called `forward` everywhere, `url` has been renamed to `dest`, and the implementation now follows the Manager -> DestHandler -> protocol-specific Dest structure you suggested.

I also reviewed the live-miracles patch and adopted its parallel shutdown approach, so thanks to @krsna1729 for that improvement.

Thanks @aler9 for moving this _forward_!

internal/forward/rtmp/client.go
```diff
@@ -0,0 +1,347 @@
+// Package rtmp contains the RTMP forward destination.
+package rtmp
```
aler9: why did you create an alternative RTMP client? If you need additional features that are not exposed by the current version of gortmplib, open a patch there. I couldn't spot any though, all i know is that this alternative client is used when "username is nil". Remove this file

internal/api/api_forward.go
```diff
@@ -0,0 +1,124 @@
...
+		} else {
+			a.writeError(ctx, http.StatusInternalServerError, err)
+		}
+		return
+	}
+
+	ctx.JSON(http.StatusOK, data)
+}
+
+func (a *API) onForwardAdd(ctx *gin.Context) {
```
aler9: this endpoint is redundant. Forward targets can already be added by using `/v3/config/paths/patch/{name}`. We don't want to provide the same feature twice. If you'd like to add single forward destinations for each API call, we can add in the future a new endpoint `/config/paths/forward/{pathName}/add`, that acts on the config, which is the source of truth. Please not in this PR.

internal/api/api_forward.go
```diff
@@ -0,0 +1,124 @@
...
+		default:
+			a.writeError(ctx, http.StatusBadRequest, err)
+		}
+		return
+	}
+
+	ctx.JSON(http.StatusOK, data)
+}
+
+func (a *API) onForwardRemove(ctx *gin.Context) {
```
aler9: this endpoint is redundant. Forward targets can already be added by using `/v3/config/paths/patch/{name}`. We don't want to provide the same feature twice

internal/api/api_forward.go
```diff
@@ -0,0 +1,124 @@
...
+	if err != nil {
+		a.writeError(ctx, http.StatusBadRequest, err)
+		return
+	}
+	data.PageCount = pageCount
+
+	ctx.JSON(http.StatusOK, data)
+}
+
+func (a *API) onForwardGet(ctx *gin.Context) {
```
aler9, line 43: if forward targets come from the config only, then they can be identified by position. UUID is redundant. this endpoint can become `/v3/paths/forward/{pathName}/get/{pos}`

internal/api/api_forward.go
```diff
@@ -0,0 +1,124 @@
...
+	"github.com/gin-gonic/gin"
+	"github.com/google/uuid"
+
+	"github.com/bluenviron/mediamtx/internal/conf"
+	"github.com/bluenviron/mediamtx/internal/conf/jsonwrapper"
+	"github.com/bluenviron/mediamtx/internal/defs"
+	"github.com/bluenviron/mediamtx/internal/forward"
+)
+
+func (a *API) onForwardList(ctx *gin.Context) {
```
aler9, line 15: this endpoint can become `/v3/paths/forward/{pathName}/list`

internal/forward/rtmp/dest_test.go
```diff
@@ -0,0 +1,41 @@
+package rtmp
```
aler9: this test file must contain a test that demonstrates that Dest is really able to connect to the intended destination and forward something. add a TestDest, with a dummy test.Parent{} that implements what's missing

internal/forward/rtsp/dest_test.go
```diff
@@ -0,0 +1,145 @@
+package rtsp
```
aler9: this test file must contain a test that demonstrates that Dest is really able to connect to the intended destination and forward something. add a TestDest, with a dummy test.Parent{} that implements what's missing

internal/forward/srt/dest_test.go
```diff
@@ -0,0 +1,120 @@
+package srt
```
aler9: this test file must contain a test that demonstrates that Dest is really able to connect to the intended destination and forward something. add a TestDest, with a dummy test.Parent{} that implements what's missing

internal/forward/dest_handler.go
```diff
@@ -0,0 +1,268 @@
...
+	PathName          string
+	Matches           []string
+	PathManager       PathManager
+	Parent            logger.Writer
+
+	ctx       context.Context
+	ctxCancel func()
+	done      chan struct{}
+
+	uuid    uuid.UUID
```
aler9: a random UUID is redundant because forward destinations are not dynamic, but precisely mapped one-to-one the configuration, which acts as source of truth. The position is the real unique identifier here, and allows to associate logs in the console to each forward destination, for instance:
"[dest 1] hello"
is clearly associated with the first entry in the `forward` list.
The alternative is using a name, but that would mean asking users for an additional configuration property, which we don't want to do right now.

internal/defs/api_forward.go
```diff
@@ -0,0 +1,64 @@
...
+	APIForwardStateError      APIForwardState = "error"
+)
+
+// APIForwardSource is where a forward was created from.
+type APIForwardSource string
+
+// forward sources.
+const (
+	APIForwardSourceConfig APIForwardSource = "config"
+	APIForwardSourceAPI    APIForwardSource = "api"
```
aler9: this must be eliminated, if we restore the architectural pattern in which the configuration is the only source of truth

aler9 (changes requested):

aler9:
I added some additional comments. The situation looks good, there is still some work to do in order to normalize a couple of things, in particular:
* the RTMP client, that must stay in gortmplib
* dynamic forward destinations through the API - the configuration, either on disk or in memory, is the source of truth. This might change in the future, but right now we can't change the current architecture.

internal/forward/rtmp/client.go
aniolpages: The issue I experimented while testing this was the YouTube was closing the connection when I tried publishing a stream. They use `rtmp://a.rtmp.youtube.com/live2` and seems that gortmplib sends `live2` as the publish type.

```
diff --git a/client.go b/client.go
@@
        err = c.mrw.Write(&message.CommandAMF0{
            ChunkStreamID:   4,
            MessageStreamID: 0x1000000,
            Name:            "publish",
            CommandID:       5,
            Arguments: []any{
                nil,
                streamKey,
-               app,
+               "live",
            },
        })

```

Changing this in gortmplib does allow publishing to YouTube, but I'm not sure if it would break any other use case of the library.

internal/forward/dest_handler.go
aniolpages: I want to confirm the expected stability semantics. Since `forward` is a configuration list, removing an entry from `[A, B, C, D, E]` produces `[A, B, D, E]`. D moves from position 4 to 3 and E from 5 to 4. Therefore, the same `get/3` request returns C before the configuration change and D afterwards.

Is this intended, with `pos` representing only the current configuration snapshot?

internal/api/api_forward.go
aniolpages: A route such as `/v3/paths/forward/:pathName/list` works for a single-segment path, but paths such as `my/nested/path` return 404, even when the slashes are encoded as `%2F`. Gin also rejects a catch-all parameter such as `*pathName` when it is followed by another route segment.

Existing MediaMTX path endpoints semm to use a trailing `*name` for this reason.

How would you like the new endpoints to represent nested paths? Would it be acceptable to keep the wildcard at the end, for example:

- `/v3/paths/forward/list/*name`
- `/v3/paths/forward/get/:pos/*name`

aniolpages:
Thanks! I did some tests and still have a few questions about using the position as the identifier.
I’m thinking about an external service managing forward destinations. If it needs to retrieve or modify a specific destination, positions don’t seem very reliable: removing one entry shifts the following ones, so the same position may later point to a different destination.

For the same reason, I’m not sure how useful `/v3/paths/forward/{pathName}/get/{pos}` would be. A client can’t be sure that the configuration hasn’t changed since it obtained the position, so it might receive a different destination from the one it expected.

This is why I originally thought UUIDs and specific endpoints for adding and removing destinations were a good idea, as they provide a stable reference and allow each operation to affect only the intended destination. These endpoints could still update the configuration internally.

aniolpages:
I understand that clients should use PATCH `/v3/config/paths/patch/{name}` and provide the complete desired forward array. Any existing destination omitted from that array would be removed. Is that correct?

This could also cause a lost-update race. For example, two clients could read `[A, B]`, then one writes `[A, B, C]` while the other writes `[A, B, D]` using its older copy. The final configuration would be `[A, B, D]`, unintentionally removing `C`.

internal/forward/rtmp/client.go
aler9: i checked in other libraries too, and that field is always set to the "app" value, which is basically the URL path. Can you perform an additional YouTube test? if you're sure that YouTube requires the `/live2` URL and the `/live` app, then we can add a field to `gortmplib.Client` called `App`:

```go
type Client struct {
	//
	// Target
	//
	// URL of the RTMP server to connect to.
	URL *url.URL
	// Whether to publish or play.
	Publish bool
	// When publishing, use this as the "app" field instead of inferring it
	App string
...
```

But it seems strange to me, i haven't seen this in other clients.

internal/api/api_forward.go
aler9: Then it's probably better to move both parameters as query arguments. The name, as you emphasized, requires dedicated space.

internal/forward/dest_handler.go
aler9: You are correct about the fact that the `pos` of an element changes when the previous one is eliminated. The root problem is that currently in the server we're not allowing to change configuration list elements one by one, except for paths. Paths are indexed by name. The advantage of using names is that there's a clear correlation between logs and their owner. Therefore it's a new thing that must be developed in order to be versatile and not linked to a specific feature.

 Therefore, available solutions are:

* using a random UUID as index - this doesn't allow to link logs to their owner
* using position as index - this doesn't allow atomic operations on forward destinations
* using name as index - this requires asking users to fill a name for each forward destination, like:
   ```yml
   forward:
     myname1:
       dest: rtmp://
     myname2:
       dest: rtmp://
   ```
   
   Probably the best solution is an hybrid approach, in which
   * logs are using `pos`
   * `pos` is still a field returned by each `Forward` entity in the API
   * Forward entities also have a `uuid` field with a random UUID, and `forward/get/{UUID}` works by using UUIDs

aler9:
Summing up:
* RTMP issue --> please perform an additional check, if it still fails send a gortmplib path.
* Forward destinations unique ID --> hybrid approach, POS in logs and UUID in API. Feel free to propose a better alternative if you have one. 
* path name and forward destination ID as API arguments --> move them both as query parameters.

internal/forward/rtmp/client.go
aniolpages: I retested this and I think we’re referring to two different fields. For a `/live2/<key>` URL, YouTube requires `connect.app = "live2"`, while the third argument of publish is the publishing type and must be `"live"`.

`connect(app="live2")` + `publish(key, "live2")` -> YouTube closes with EOF
`connect(app="live2")` + `publish(key, "live")` -> NetStream.Publish.Start

In the 7.2.2.6 section of the RTMP specification, publishing type can be "live", "record" or "append". In this case I think w're wrongly using the app name as the publishing type.

internal/forward/dest_handler.go
aniolpages: My only concern is using `pos` alone in logs: after removing or reordering an entry, the same position can refer to different destinations in historical logs. Could we include both the current position and a shortened UUID in the log prefix, for example `[dest 2 a1b2c3d4]`? 

The API and metrics would use the full UUID, while `pos` would only represent the current configuration order.

aniolpages:
I’d move the path name and destination ID to query parameters, and use the hybrid approach with the full UUID in the API and the current position in the response.
For logs I’d suggest including both the position and a shortened UUID, for example `[dest 2 a1b2c3d4]`, since positions can shift and otherwise the same log prefix could refer to different destinations over time.

About RTMP / YouTube, the additional tests confirm that the problem is the publish type argument, not the app value. YouTube URL and app remain `live2`, but the publish command needs the type `live`.
Should I open a PR in gortmplib adding an optional PublishType field?

aler9:
> Should I open a PR in gortmplib adding an optional PublishType field?

<s>Yes, open a PR there, thanks</s>

Edit: i created it directly https://github.com/bluenviron/gortmplib/pull/95

internal/forward/dest_handler.go
aler9: > Could we include both the current position and a shortened UUID in the log prefix, for example [dest 2 a1b2c3d4]?

yes, this is OK

internal/forward/rtmp/dest.go
```diff
@@ -0,0 +1,152 @@
...
+		fourCCToString(message.FourCCAVC),
+		fourCCToString(message.FourCCOpus),
+		fourCCToString(message.FourCCFLAC),
+		fourCCToString(message.FourCCAC3),
+		fourCCToString(message.FourCCMP4A),
+		fourCCToString(message.FourCCMP3),
+	}
+}
+
+func urlWithDefaultPort(u *url.URL) *url.URL {
```
aler9: concidentally this has just been added to gortmplib natively (see https://github.com/bluenviron/gortmplib/pull/99) so it can be removed

internal/forward/rtmp/dest_test.go
```diff
@@ -0,0 +1,189 @@
...
+		}},
+	}}}))
+
+	require.NotEmpty(t, fourCCList(&description.Session{Medias: []*description.Media{{
+		Type:    description.MediaTypeVideo,
+		Formats: []format.Format{&format.H265{}},
+	}}}))
+}
+
+func TestURLWithDefaultPort(t *testing.T) {
```
aler9: concidentally this has just been added to gortmplib natively (see https://github.com/bluenviron/gortmplib/pull/99) so it can be removed

aler9 (changes requested):

aniolpages:
Thanks for the improvements!

It looks like a UUID is now only preserved when the destination remains at the same position. If an entry is inserted or removed, all shifted destinations get new UUIDs. The old UUIDs return 404 instead of pointing to another destination, which is good, but this still means an external controller cannot use them as stable identifiers across configuration changes.

Is this the intended behaviour? Would it make sense for the UUID to be part of the destination configuration, so it remains stable when entries are added, removed or reordered?

aler9:
@aniolpages the problem about the original hot reloading feature is that it used the "dest" string as unique key, while in the future forward destinations will be certainly populated with additional fields, for instance:

```yml
forward:
- dest: rtsp://...
  rtspTransport: tcp

- dest: whep://
  whepTokenBearer: xxx
```

Therefore, that was not technically possible. What is possible and can be added in a future patch is using the configuration fingerprint (`sha256(ForwardDest)`) as unique identifier in order not to lose any previous UUID.

aler9 (approved):

aler9:
Thank you very much for all this work - it was long to review and adapt because it needs to integrate with the existing ecosystem, and there were also lots of small details that were merged through dedicated patches.

This can also be the foundation for a huge amount of improvements, from additional protocols (WebRTC, MoQ) to additional parameters (rtspTransport, bearer token, etc) - i avoided contributing additional features with respect to the original patch because it was already quite vast for a single PR.

I'm also expecting some bugs to appear in the following weeks or months - not because of this patch, but because it will be the first time that all the sub-libraries (gosrt, gortmplib, gortsplib) will be used to publish by a massive amount of users and towards a great variety of servers and devices. Let's see what happens.

PR #5569, opened 2026-03-13
feat: handle iss and aud claims for JWT auth

MrLightful:
Enhances MediaMTX JWT authentication by validating the iss (issuer) and aud (audience) claims. Tokens are now rejected if they are not issued by a trusted authority or not intended for this service, preventing misuse across APIs.

The claims allow MediaMTX to enforce the recommended security practice, as specified in RFC 7519 and OpenID Connect, preventing token misuse and cross-service attacks.

The claims are optional. The validation is skipped if options are empty in the config. Thus, is not a breaking change.

### DIFF
--- a/api/openapi.yaml
+++ b/api/openapi.yaml
@@ -122,6 +122,10 @@ components:
           type: string
         authJWTClaimKey:
           type: string
+        authJWTIssuer:
+          type: string
+        authJWTAudience:
+          type: string
         authJWTExclude:
           type: array
           items:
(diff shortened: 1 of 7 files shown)


MrLightful:
@aler9 if I get it right, the test isn't failing because of my changes, right? I see other feature branches failing as well with this test/error.

MrLightful:
Rebased branch with main. Not sure if that will help the test.

aler9:
yes i saw the test failure, it's unrelated. I'll merge this as soon as possible. Thanks!

aler9:
merged, thanks!

PR #5577, opened 2026-03-16
docs: clarify jwt in query option for rtsp and rtmp

MrLightful:
I was going through the code while working on #5569 and noticed that RTSP and RTMP always allow jwt in http query, regardless of the setting in the config. It wasn't very obvious to me, so suggest a bit of clarification in the config example.

Reasoning: I was having it enabled, because thought was required for RTSP. But it wasn't, and just introduces additional security implications for other protocols that don't need it.


aler9:
Hello, yes, the comment describes exactly what the situation is. Until major clients keep legacy and old restrictions on password sizes, there's nothing we can do.

PR #5592, opened 2026-03-19
core: add reloadOnError option to survive configuration errors

aries-sd:
When a configuration file change triggers a reload and the new configuration contains a parse or apply error, the current behavior is to exit the process. This kills all active streams, potentially causing an avoidable outage.

This adds a new "reloadOnError" configuration option:
- "exit" (default): exit the process (current behavior, unchanged)
- "continue": log the error and keep running with the current configuration

The option only affects hot reloads. A broken configuration at initial startup always exits, since there is no previous configuration to fall back to.

### DIFF
--- a/internal/conf/conf.go
+++ b/internal/conf/conf.go
@@ -247,6 +247,7 @@ type Conf struct {
 	LogFile             string          `json:"logFile"`
 	SysLogPrefix        string          `json:"sysLogPrefix"`
 	DumpPackets         bool            `json:"dumpPackets"`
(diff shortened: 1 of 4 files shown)


aler9:
Hello, currently the server can be "up and running" or "terminated due to an invalid configuration". This introduces a third state which is "up and running with a configuration that is not clearly visible on disk anymore", which is something that most users do not want and might create confusion.

An alternative that would be welcome would be implementing a `--check` flag which makes the server to verify the configuration file and then exit.

PR #5604, opened 2026-03-22
Add support for KLV in HLS

EricHenry:
Adds support for KLV tracks in HLS streams. Using https://github.com/bluenviron/gohlslib/pull/337 

Note: will update and delete go.mod when above is merged.


aler9:
@EricHenry can you update this patch and remove the go.mod change?

Also, KLV in HLS works only when variant is MPEG-TS, so can you edit `from_stream.go` and add the KLV track only if the variant is MPEG-TS?

Thanks

EricHenry:
@aler9 thanks for feedback! Updated with the suggestions. Let me know if there is anything I am missing. :+1:

aler9:
finally merged, thanks for all the work!

PR #5663, opened 2026-04-13
Metrics string builder

JohanG-LAS:
Replace string concatenation with strings.Builder in the metrics handler.

Benchmark results (full handler, all protocols including SRT):

Time: 590 us → 179 us (-70%)
Memory: 1,462 KB → 114 KB (-92%)
Allocations: 627 → 298 (-52%)


aler9:
i applied linter and merged, thank you very much!

PR #5672, opened 2026-04-17
Allow configuration of behaviour if each server fails to start

kelvinlawson:
On a production system, you might prefer to allow the system to at least start with the mandatory ports running and not completely shut down if one inconsequential server fails to start because the port is in use.

### DIFF
--- a/internal/conf/conf.go
+++ b/internal/conf/conf.go
@@ -273,126 +273,137 @@ type Conf struct {
 	AuthJWTAudience           string                       `json:"authJWTAudience"`
 
 	// Control API
-	API               bool       `json:"api"`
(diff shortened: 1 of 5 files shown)


kelvinlawson:
Defaults to the current behaviour (all services are set to "fatal" and MediaMTX will shut down if they fail to start).

aler9:
Hello, currently the server can be either "up and running" or "not running". This introduces a third state which is "running even when some listeners fails" which is not supported by any server i know of and doesn't allow to immediately identify configuration errors. The server is intended to fail if the configuration is wrong or if it cannot create every enabled listener. If you have troubles, disable unused listeners in the configuration file.

PR #5707, opened 2026-04-29
prevent code injection in case of MTX_QUERY in hooks

aler9:
When MTX_QUERY is used explicitly in hooks, for instance "curl http://something/?$MTX_QUERY", it can be used to inject arbitrary commands. MTX_QUERY is now url-encoded to prevent any abuse regardless of the configuration.

### DIFF
--- a/docs/2-features/21-hooks.md
+++ b/docs/2-features/21-hooks.md
@@ -58,7 +58,7 @@ pathDefaults:
   # This is terminated with SIGINT when there are no readers anymore.
   # The following environment variables are available:
   # * MTX_PATH: path name
-  # * MTX_QUERY: query parameters (passed by first reader)
+  # * MTX_QUERY: query parameters (passed by first reader) (url-encoded)
   # * RTSP_PORT: RTSP server port
   # * G1, G2, ...: regular expression groups, if path name is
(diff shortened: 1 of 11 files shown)


McNutnut:
After this change, when using something like,

`runOnDemand: /usr/bin/ffmpeg -i https://my-upstream.com/?$MTX_QUERY`

and a client supplying,

`?my-upstream-key=ABC&my-upstream-type=XYZ`, 

how would I go about passing this to the `runOnDemand` without the query being encoded?  Such as `https://my-upstream.com/?my-upstream-my-upstream-key=ABC&my-upstream-type=XYZ` and not `https://my-upstream.com/?my-upstream-key%3DABC%26my-upstream-type%3DXYZ`

aler9:
@McNutnut sorry for the breaking change but it was technically impossible to keep $MTX_QUERY unsanitized and prevent arbitrary command injection - anyway, you can restore your integration and at the same time make it secure by using printf to decode MTX_QUERY:

```yml
runOnDemand: sh -c '/usr/bin/ffmpeg -i "https://my-upstream.com(printf "%%b" "${MTX_QUERY//%%/\\x}")"'
```

Issue #5718, opened 2026-04-30
[Feature Request] On-Demand Dynamic Path Creation via External API


mandeerson:
### Describe the feature

**Description**
I would like to propose a feature that enables dynamic, on-demand creation of paths when a client requests a stream that is not yet configured in MediaMTX.

**Use Case**
In some scenarios, stream sources and their configurations are not known in advance or are managed externally (e.g., in a database or service). Currently, all paths must be predefined in the MediaMTX configuration. This creates limitations for systems that need to scale dynamically or integrate with external orchestration logic.

**Proposed Behavior**

1. When a client requests a path that does not exist:
2. MediaMTX detects that the requested path is not configured.
3. Instead of immediately returning an error or all_others, MediaMTX triggers a request to a configurable external API endpoint.
4. This API returns the necessary configuration for that path (e.g., source, authentication, parameters).
5. MediaMTX dynamically creates the path using the returned data.
6. The stream is then initiated and served to the client.

**Example Flow**

1. Client requests: /camera123
2. Path is not found in MediaMTX
3. MediaMTX sends a request to: POST /resolve-path (external API)
4. API responds with path configuration (e.g., source URL, credentials)
5. MediaMTX creates /camera123 dynamically
6. Stream starts and is delivered to the client

This feature would greatly improve flexibility for dynamic streaming environments and allow MediaMTX to integrate more easily into modern, service-oriented architectures.

khakra:
+100, requested something similar here https://github.com/bluenviron/mediamtx/issues/5321. This will be extremely helpful.

kush1jpeg:
much needed feat....any1 working on this currently?

michelreipsporter:
I created a pull request adding this functionality.

aler9:
This feature is already provided by the `runOnDemand` hook, that is called whenever its corresponding path is requested:

```yml
paths:
  all:
    runOnDemand: http://my-backend/create-path?path=$MTX_PATH
```

You can use `runOnDemand` to point to a backend that makes the requested stream available.

Loading configuration through an external HTTP endpoint is something that would mess up lots of things, in particular the hot reloading feature, so we can't do it directly right now.

PR #5734, opened 2026-05-05
feat: adding support to dynamic external path load

michelreipsporter:
https://github.com/bluenviron/mediamtx/issues/5718


fabiankoebel:
Thanks for the updates and the effort you've put into this PR.

Could you please upmerge the latest `main` branch again and resolve the remaining conflicts? Once that's done, is there a way to request integration into `main`, or is there any additional process we should follow?

Thanks!

michelreipsporter:
Thanks for the review.

I've already upmerged the latest main branch and resolved all remaining conflicts. The branch is now up to date with main and should be ready for integration.

Regarding the integration, the implementation included in this PR should be enough. To use it, it is only necessary to configure the external endpoint and enable the feature through:

```yaml
pathExternalConfEnabled: true
pathExternalConfURL: http://your-server/api/streams
```

Once those settings are configured, MediaMTX will fetch path configurations from the external endpoint and the integration should work as expected.

Please let me know if there's any additional process required before merging.
Thanks!

aler9:
https://github.com/bluenviron/mediamtx/issues/5718#issuecomment-5117792506

PR #5753, opened 2026-05-12
Add user agent support for RTMP, RTSP, WebRTC, and HLS

alexmck:
## Summary

Adds user agent reporting for RTMP, RTSP, WebRTC, and HLS sessions. This gives visibility into what software clients are using to connect, which is useful for debugging compatibility issues and monitoring ingest sources.

 - **RTMP**: Exposes `flashVer` from the RTMP handshake as the user agent. Depends on an upstream gortmplib change (https://github.com/bluenviron/gortmplib/pull/68) to expose `flashVer` on ServerConn.
 - **RTSP**: Reads the `User-Agent` header from the ANNOUNCE request.
 - **WebRTC**: Reads the `User-Agent` header from the HTTP request.
 - **HLS**: Reads the `User-Agent` header from the HTTP request.

  The user agent string is included in:
 - The **API** response for each protocol as a `userAgent` field:
   - RTMP connections (`/v3/rtmpconns/list`)
   - RTSP sessions (`/v3/rtspsessions/list`)
   - WebRTC sessions (`/v3/webrtcsessions/list`)
   - HLS sessions (`/v3/hlssessions/list`)
 - The **authentication webhook** payload as a `userAgent` field
 - **Log output** on publish, when a user agent is present

I also considered adding a `MTX_USER_AGENT` environment variable that could be used in other webhooks, but decided against it. If this is something that you feel could be useful, this is something that could be added in a follow up PR.

This is a non-breaking change. I have also tested this PR in production and I can confirm that it works as expected.

## SRT

SRT does not have a user agent field.


aler9:
This is OK, you are missing HLS, that is session-based since v1.18.1 and can return user-agents too.

alexmck:
Thank you. I have added this change to HLS as requested. And thank you for accepting the other PR on gortmplib. Once gortmplib has a new tag, I can update this PR to include that and then this PR will be ready to merge.

alexmck:
Thank you for tagging a new release of gortmplib and bumping that dependency on metdiamtx. This feature is now ready for merging.

alexmck:
I've just pushed an update to address those failing tests due to the addition of the userAgent in the API response.

aler9:
i added a missing case (RTSP readers) and merged, thanks!

alexmck:
Thank you!

Issue #602, opened 2021-09-30
Support PROXY protocol when running behind Load Balancer


aler9:
### Discussed in https://github.com/aler9/rtsp-simple-server/discussions/477

<div type='discussions-op-text'>

<sup>Originally posted by **dorinclisu** July 27, 2021</sup>
Is it difficult to support it? https://www.haproxy.org/download/1.8/doc/proxy-protocol.txt

The reason is, behind a load balancer all rtsp server request IP's are of the balancer itself, therefore losing the ability to whitelist client IP's.

The load balancer I'm using does have a proxy toggle, and specifically warns it that the application must support it if enabled. I did try it on and the request wouldn't reach the rtsp server anymore so I assume it doesn't support it yet.</div>

kuolemaaa:
any news on this? is it supported? or set aside?

aler9:
implemented with #5754

PR #5754, opened 2026-05-12
feat: Add PROXY protocol support for RTMP and RTSP servers

alexmck:
## Summary

Adds support for PROXY protocol v1/v2 for RTMP, RTMPS, RTSP, and RTSPS ingest, allowing the real client IP to be preserved when connections pass through a load balancer or reverse proxy. This feature was first mentioned in #602.

Two new configuration options control this:
  - `rtspTrustedProxies` list of IPs/CIDRs trusted to send PROXY headers on RTSP connections
  - `rtmpTrustedProxies` same for RTMP connections
  
This follows the same format that `webrtcTrustedProxies` uses.

Connections from trusted proxies have their PROXY protocol header parsed, and the original client IP is used for logging and authentication. Connections from non-trusted IPs ignore any PROXY header and are passed through unchanged.

For RTMPS, the PROXY header is read before TLS negotiation to ensure correct ordering.

Also, when `rtspTrustedProxies` is configured with RTSPS encryption, the `dumpPackets` feature will not capture TLS-level traffic for RTSP. I assume the undocumented `dumpPackets` is for debug only, but wanted to mention it just in case this is intentionally here. I can make a revision here if requested.

This is a non-breaking change. I have also tested this PR in production and I can confirm that it works as expected.

## New Dependency

This PR uses the [go-proxyproto](https://github.com/pires/go-proxyproto) library. I wanted to make sure it's very clear that this PR introduces a new dependency to the library.

Please let me know if you have any questions.


internal/servers/rtsp/server.go
```diff
@@ -187,6 +190,21 @@ func (s *Server) Initialize() error {
...
+			innerListen = net.Listen
+		}
+		s.srv.Listen = func(network, address string) (net.Listener, error) {
+			ln, err := innerListen(network, address)
+			if err != nil {
+				return nil, err
+			}
+			return proxyprotocol.WrapListener(ln, s.TrustedProxies), nil
+		}
+		s.srv.TLSListen = nil
```
aler9: this prevents s.DumpPackets from working. Find a solution to make both work

internal/servers/rtmp/server.go
```diff
@@ -142,9 +155,18 @@ func (s *Server) Initialize() error {
...
-		s.ln, err = tlsListen(net, addr, &tls.Config{GetCertificate: s.loader.GetCertificate()})
-		if err != nil {
-			return err
+
+		if len(s.TrustedProxies) > 0 {
+			s.ln, err = listen(net, addr)
+			if err != nil {
+				return err
+			}
+			s.ln = tls.NewListener(s.ln, &tls.Config{GetCertificate: s.loader.GetCertificate()})
```
aler9: this prevents s.DumpPackets from working. Find a solution to make both work

internal/protocols/proxyprotocol/listener.go
```diff
@@ -0,0 +1,25 @@
+package proxyprotocol
```
aler9: move the package in internal/protocols/proxyprotocol

internal/core/core.go
```diff
@@ -474,6 +474,7 @@ func (p *Core) createResources(initial bool) error {
 			ServerCert:          "",
 			ServerKey:           "",
 			RTSPAddress:         p.conf.RTSPAddress,
+			TrustedProxies:      p.conf.RTSPTrustedProxies,
```
aler9, line 494: add these new parameters to openapi.yaml and check that all linters pass

aler9 (changes requested):
thanks for providing this long-awaited feature, i left some comments. In particular, we must find a way to make dumpPackets and trustedProxies coexist.

alexmck:
Thank you for your review and feedback on this PR. I now have addressed all feedback items and lint issues. 

I have made one tradeoff so the PROXY header bytes themselves don't appear in the packet dump, since they're stripped before the packetdumper sees the connection. Adding these headers to the dump would require a change to the packetdumper. I feel this could be achieved in a follow up PR if requested, but I gather the main reason for this feature is inspecting the media stream which this change will still allow.

Please let me know if you have any other feedback or concerns.

aler9 (approved):

aler9:
i had to perform a significant refactor due to the existing packetdumper architecture, but in the end i was able to merge. Thanks for providing the original implementation.

alexmck:
Thank you for your fixes and refactor. It's much appreciated.

PR #5759, opened 2026-05-12
Add max age to HLS CORS preflight responses

### LINKED ISSUE #5755: Add Access-Control-Max-Age header to reduce OPTIONS requests
### Describe the feature

Setting `Access-Control-Max-Age: 86400` for example can reduce `OPTIONS` requests from browsers.

Now `OPTIONS` is fetched before every HSL segment:

Image

modeitsch:
Fixes #5755.

This adds `Access-Control-Max-Age: 86400` to HLS CORS preflight responses so browsers can cache successful preflight checks instead of sending an OPTIONS request before every segment request.

The existing HLS preflight test now verifies the new header.

Tests:
- `go test ./internal/servers/hls -run TestServerPreflightRequest -count=1`
- `go test ./internal/servers/hls -count=1`


aler9:
Hello, while caching CORS preflight responses would be extremely useful, in reality with Low-latency HLS it is not technically possible, since `Access-Control-Max-Age` uses path and query as cache key, and in LL-HLS the path and query are different for every request:

Screenshot_20260620_153719

Therefore nothing is cached. Unfortunately.

PR #5762, opened 2026-05-13
Add community-maintained webrtc reader for react to docs

MrLightful:
The existing reader.js was somewhat annoying to use with react, so I ported it:
https://github.com/MrLightful/mediamtx-webrtc-react

Don't want for my pr to look too spammy, but I imagined it could be useful for others. 

The package provides multiple approaches of handling webrtc streams, incl. simple component based, hooks, or class based (similar to original reader.js).

### DIFF
--- a/docs/4-read/13-web-browsers.md
+++ b/docs/4-read/13-web-browsers.md
@@ -108,6 +108,57 @@ After the video tag, add a script that initializes the stream when the page is f
 </script>
 ```
 
+### WebRTC with React
+
+The community-maintained [mediamtx-webrtc-react](https://www.npmjs.com/package/mediamtx-webrtc-react) package provides React hooks and components for reading WebRTC streams from _MediaMTX_.
+
+Install the package:
+
+```bash
+npm install mediamtx-webrtc-react
+```
+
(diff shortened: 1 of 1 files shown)


aler9:
Hello, short history of the WebRTC/WHEP JS reader: i never wanted to write one and i wanted to rely on a community-maintained library like hls.js. Turned out that there were none, so i had to write it from scratch.

The current JS reader is designed to minimize maintenance efforts, since it doesn't require a JavaScript transcoder, and is also designed to maximize compatibility, since it's not bound to a specific framework. Furthermore, it's constantly updated for various reasons, and these updates must reach all users ASAP. The main issue with a community-maintained React layer is that it delays updates.

However, i realize that using a traditional library in the React world is difficult for some users, even though writing a wrapper is not difficult.

There's a solution that would met all requirements (no transcoders, compatibility with any framework, ease of use with reactive frameworks), that is rewriting the JS reader as a Web Component, which is React-friendly and natively supported by browsers and other frameworks. The resulting work can be placed in this repository and replace the JS class, in order to maintain a single source of truth. We're open to that.

PR #5767, opened 2026-05-14
always-available: support disabling recording when source is offline

### LINKED ISSUE #5587: Add option to prevent 24/7 recording when alwaysAvailable is enabled
### Describe the feature

When the alwaysAvailable flag is enabled, MediaMTX begins recording 24/7 even if no source is currently connected to the path. This leads to continuous creation of empty or useless recordings.

It would be very helpful to have an additional option—something like recordOnConnect—that ensures recording only starts when an actual source connects (i.e., when the path becomes active due to an incoming stream, not simply because it is marked as always available).

This would allow users to keep paths pre-initialized with alwaysAvailable while avoiding unnecessary recordings when no stream is present.

AbdElrahmaN31:
## Summary

Add `alwaysAvailableRecorded` parameter that controls whether recording happens during offline periods for always-available paths. When set to `false`, recording only starts when a real source is connected and stops when it disconnects.

Default is `true`, preserving current behavior.

Closes #5587

## Changes

- `internal/conf/path.go`: Add `AlwaysAvailableRecorded` config field (default: `true`)
- `internal/core/path.go`: Conditional recording start/stop based on the new flag in `setAvailable()`, `doAddPublisher()`, `executeRemovePublisher()`, and `doReloadConf()`
- `internal/conf/conf_test.go`: Update test expected struct
- `docs/2-features/08-always-available.md`: Document the new parameter

## Testing

- All config tests pass (`go test ./internal/conf/...`)
- Core tests pass (pre-existing SRT flaky test is unrelated)
- Build succeeds (`go build ./...`)

### DIFF
--- a/docs/2-features/08-always-available.md
+++ b/docs/2-features/08-always-available.md
@@ -26,3 +26,15 @@ paths:
     alwaysAvailable: true
     alwaysAvailableFile: "./h264.mp4"
 ```
+
+By default, the stream is recorded even when no source is connected and the offline segment is playing. To record only when a source is connected, set `alwaysAvailableRecorded` to `false`:
+
+```yml
+paths:
+  mypath:
+    alwaysAvailable: true
+    alwaysAvailableRecorded: false
+    record: true
+    alwaysAvailableTracks:
+      - codec: H264
+```
--- a/internal/conf/conf_test.go
+++ b/internal/conf/conf_test.go
@@ -48,6 +48,7 @@ func TestConfFromFile(t *testing.T) {
 			SourceOnDemandCloseAfter:     10 * Duration(time.Second),
 			OverridePublisher:            true,
 			AlwaysAvailableTracks:        []AlwaysAvailableTrack{},
+			AlwaysAvailableRecorded:      true,
 			RecordPath:                   "./recordings/%path/%Y-%m-%d_%H-%M-%S-%f",
 			RecordFormat:                 RecordFormatFMP4,
 			RecordPartDuration:           Duration(1 * time.Second),
--- a/internal/conf/path.go
+++ b/internal/conf/path.go
@@ -193,9 +193,10 @@ type Path struct {
 	UseAbsoluteTimestamp       bool     `json:"useAbsoluteTimestamp"`
 
 	// Always available
-	AlwaysAvailable       bool                   `json:"alwaysAvailable"`
-	AlwaysAvailableTracks []AlwaysAvailableTrack `json:"alwaysAvailableTracks"`
-	AlwaysAvailableFile   string                 `json:"alwaysAvailableFile"`
+	AlwaysAvailable         bool                   `json:"alwaysAvailable"`
+	AlwaysAvailableTracks   []AlwaysAvailableTrack `json:"alwaysAvailableTracks"`
+	AlwaysAvailableFile     string                 `json:"alwaysAvailableFile"`
+	AlwaysAvailableRecorded bool                   `json:"alwaysAvailableRecorded"`
 
 	// Record
 	Record                bool         `json:"record"`
@@ -319,6 +320,9 @@ func (pconf *Path) setDefaults() {
 	pconf.Source = "publisher"
 	pconf.SourceOnDemandStartTimeout = 10 * Duration(time.Second)
 	pconf.SourceOnDemandCloseAfter = 10 * Duration(time.Second)
+
+	// Always available
+	pconf.AlwaysAvailableRecorded = true
 
 	// Record
 	pconf.RecordPath = "./recordings/%path/%Y-%m-%d_%H-%M-%S-%f"
--- a/internal/core/path.go
+++ b/internal/core/path.go
@@ -404,12 +404,14 @@ func (pa *path) doReloadConf(newConf *conf.Path) {
 			newConf.RecordPartDuration != oldConf.RecordPartDuration ||
 			newConf.RecordMaxPartSize != oldConf.RecordMaxPartSize ||
 			newConf.RecordSegmentDuration != oldConf.RecordSegmentDuration ||
-			newConf.RecordDeleteAfter != oldConf.RecordDeleteAfter) {
+			newConf.RecordDeleteAfter != oldConf.RecordDeleteAfter ||
+			newConf.AlwaysAvailableRecorded != oldConf.AlwaysAvailableRecorded) {
 		pa.recorder.Close()
 		pa.recorder = nil
 	}
 
-	if newConf.Record && pa.stream != nil && pa.recorder == nil {
+	if newConf.Record && pa.stream != nil && pa.recorder == nil &&
+		(!newConf.AlwaysAvailable || newConf.AlwaysAvailableRecorded || pa.source != nil) {
 		pa.startRecording()
 	}
 }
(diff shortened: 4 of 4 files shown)


xodbs1021 (commented):
Hi @AbdElrahmaN31, thanks for this PR — this switch is exactly what's needed when the offline segment acts as a standby slate that must not reach storage. I tested it in depth for a publisher-path use case, so here are the results plus two gaps I found on static-source paths, with a suggestion that closes both.

**Verified on publisher paths** (branch rebased locally onto current `main` at e175003 — see the rebase note below):

- With `alwaysAvailable: yes` + `record: yes` + `alwaysAvailableRecorded: false`, nothing is recorded while only the offline segment plays; recording starts on publisher connect and stops on disconnect, across repeated RTSP and SRT sessions.
- No offline frames leak into recordings at either boundary (every recorded file ffprobes as publisher content only). That matches the code: `SubStream.Initialize()` closes the offline substream synchronously before `startRecording()` runs, and on disconnect the recorder is closed before `StartOfflineSubStream()`.
- The final partial segment is flushed on disconnect and `runOnRecordSegmentComplete` fires for it.
- Paths added at runtime via `POST /v3/config/paths/add` honor the flag as well.
- `go build ./...` and the conf/stream/core record tests pass on the rebased branch.

**Rebase needed.** The branch currently conflicts with `main` (v1.18-era base) — 4 small spots, caused by the `setOnline()`/`setOffline()` refactor (#5956), the `runOnReady` → `runOnAvailable` rename (#5957) and new conf fields. The logic ports cleanly.

**Gap 1 — static-source ready transitions don't start recording.** `doAddPublisher()` compensates for publishers, but `doSourceStaticSetReady()` has no equivalent (and `doSourceStaticSetNotReady()` no recorder shutdown). With `source: rtsp://…` + `alwaysAvailable: yes` + `record: yes` + `alwaysAvailableRecorded: false`, the path goes `"online": true` yet no recording starts for the whole online window.

**Gap 2 — a config reload while offline turns slate recording on (static paths).** The restart gate in `doReloadConf()` uses `pa.source != nil` as an "online" proxy, but for static paths `pa.source` holds the `staticsources.Handler` from path startup, even while the upstream is disconnected. In my run, a `PATCH /v3/config/paths/patch/{name}` with just `{"recordDeleteAfter": "48h"}` while offline made offline-segment files appear within seconds, and they kept rotating until shutdown (ready/not-ready transitions don't stop it, because of gap 1). On your current base there was no better predicate, but after rebasing, `pa.isOnline()` expresses exactly the intended condition.

**Suggestion: after rebasing, move the start/stop into `setOnline()`/`setOffline()`.** Those are called for publishers and static sources alike, so both gaps close, and the two source-transition sites in `doAddPublisher()`/`executeRemovePublisher()` collapse into the shared methods:

```go
func (pa *path) setOnline(sourceDesc *defs.APIPathSource, publisherQuery string) {
	// ...existing body...
	pa.onlineTime = time.Now()

	if pa.conf.AlwaysAvailable && pa.conf.Record &&
		!pa.conf.AlwaysAvailableRecorded && pa.recorder == nil {
		pa.startRecording()
	}
}

func (pa *path) setOffline() {
	if pa.conf.AlwaysAvailable && !pa.conf.AlwaysAvailableRecorded && pa.recorder != nil {
		pa.recorder.Close()
		pa.recorder = nil
	}

	if pa.onOfflineHook == nil {
		return
	}

	pa.onOfflineHook()
	pa.onOfflineHook = nil
}
```

plus `pa.isOnline()` instead of `pa.source != nil` in the `doReloadConf()` restart gate (the `setAvailable()` condition and the reload restart-param list from this PR stay as they are). Non-always-available paths are unaffected: `setAvailable()` also calls `setOnline()` for them, but the new block is gated on `pa.conf.AlwaysAvailable`, so their recording keeps starting from `setAvailable()` as today. I applied exactly this on top of the rebase and verified it: one regression test per gap fails on the current shape and passes with the centralized one, the existing record-related core tests plus `./internal/conf` and `./internal/stream` stay green, and the publisher-path behavior above is unchanged.

**Two smaller things:**

- The new parameter is missing from the sample `mediamtx.yml` (the other `alwaysAvailable*` params are under "Default path settings -> Always available").
- A behavior test would keep this from regressing — `TestPathRecord` already has the skeleton. I wrote three along those lines (offline/connect/disconnect on a publisher path, plus one regression per gap above — the gap ones fail on the current shape, so they double as runnable reproductions); the publisher one is below, and I'm happy to share all of them together with the rebase + fix as a diff, or as a follow-up PR to your branch if that's easier.

<details>
<summary>Proposed publisher-path test</summary>

```go
func writeH264TestPacket(c *gortsplib.Client, medi *description.Media, seq uint16, ts uint32) error {
	return c.WritePacketRTP(medi, &rtp.Packet{
		Header: rtp.Header{
			Version:        2,
			Marker:         true,
			PayloadType:    96,
			SequenceNumber: seq,
			Timestamp:      ts,
			SSRC:           563423,
		},
		Payload: []byte{5},
	})
}

func waitForFileCount(t *testing.T, dir string, count int, timeout time.Duration) []os.DirEntry {
	deadline := time.Now().Add(timeout)
	for {
		files, err := os.ReadDir(dir)
		if err == nil && len(files) >= count {
			return files
		}
		if time.Now().After(deadline) {
			t.Fatalf("timed out waiting for %d file(s) in %s", count, dir)
		}
		time.Sleep(50 * time.Millisecond)
	}
}

func dirCountAndSize(t *testing.T, dir string) (int, int64) {
	files, err := os.ReadDir(dir)
	require.NoError(t, err)
	var size int64
	for _, f := range files {
		fi, err2 := os.Stat(filepath.Join(dir, f.Name()))
		require.NoError(t, err2)
		size += fi.Size()
	}
	return len(files), size
}

func TestPathRecordAlwaysAvailableRecorded(t *testing.T) {
	dir := t.TempDir()

	p, ok := newInstance(t,
		"record: yes\n"+
			"recordPath: "+filepath.Join(dir, "%path/%Y-%m-%d_%H-%M-%S-%f")+"\n"+
			"paths:\n"+
			"  mystream:\n"+
			"    alwaysAvailable: yes\n"+
			"    alwaysAvailableRecorded: false\n"+
			"    alwaysAvailableTracks:\n"+
			"      - codec: H264\n")
	require.Equal(t, true, ok)
	defer p.Close()

	// while offline, the offline substream must not be recorded
	time.Sleep(1 * time.Second)
	_, err := os.ReadDir(filepath.Join(dir, "mystream"))
	require.True(t, os.IsNotExist(err))

	// while a publisher is connected, recording must run
	media0 := test.UniqueMediaH264()

	source := gortsplib.Client{}
	err = source.StartRecording(
		"rtsp://localhost:8554/mystream",
		&description.Session{Medias: []*description.Media{media0}})
	require.NoError(t, err)

	for i := range 4 {
		err = writeH264TestPacket(&source, media0, 1123+uint16(i), 45343+90000*uint32(i))
		require.NoError(t, err)
	}

	waitForFileCount(t, filepath.Join(dir, "mystream"), 1, 3*time.Second)

	// after the publisher disconnects, recording must stop:
	// no new files may appear and the finalized file must not keep growing
	source.Close()
	time.Sleep(1 * time.Second)

	n1, size1 := dirCountAndSize(t, filepath.Join(dir, "mystream"))
	require.Equal(t, 1, n1)

	time.Sleep(1 * time.Second)

	n2, size2 := dirCountAndSize(t, filepath.Join(dir, "mystream"))
	require.Equal(t, 1, n2)
	require.Equal(t, size1, size2)
}
```

</details>

Really useful feature — happy to help get it over the line.

xodbs1021:
Following up on my review, I rebased this onto main and opened #6182 with your commit kept as the first commit; while testing it I found two gaps (a static source never started recording, and a reload while the source was offline recorded the offline segment) and covered them in a follow-up commit. If you would rather continue this PR yourself, tell me and I will close mine.

aler9:
i agree with the analysis performed by @xodbs1021, this does not cover static sources. Moved to #6182 (i will give credits to @AbdElrahmaN31 too anyway).

PR #5788, opened 2026-05-20
webrtc: add ability to exclude interfaces

tuomoart:
Adds the ability to automatically use all interfaces except some for WebRTC PeerConnection host candidates by adding a new configuration parameter `webrtcIPsFromInterfacesExcludeList`. This is useful for example for only excluding VPN interfaces while allowing all others.

Explicitly defining interfaces to use using `webrtcIPsFromInterfacesList` can be cumbersome for excluding because interface names can vary between machines and change.

### DIFF
--- a/api/openapi.yaml
+++ b/api/openapi.yaml
@@ -553,6 +553,10 @@ components:
           type: array
           items:
             type: string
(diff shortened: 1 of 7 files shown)


aler9:
i rebased and merged, thanks

PR #5790, opened 2026-05-20
Replace embedded version getter with Go build info for version

maruel:
internal/core/versiongetter/main.go: This utility used go-git to retrieve the version from the repository tags at go generate. Replace it with Go's debug.ReadBuildInfo() (added in go1.12), which provides version and VCS metadata from the binary itself.

scripts/binaries.mk: use variables to shell out less often. Switch from using ./VERSION to using the git tag. This is likely the most tricky part of this change.

scripts/dockerhub.mk already uses `git describe --tags | tr -d v` as the version. So this does help make both docker version and tarballs versioning more coherent. Note that it's different from what scripts/binaries.mk used. I didn't change the command in script/binaries.mk.

Behavior change:

The local go install automatically bumps the version for build with **modified local
changes**:

```
go generate ./... && go install . && mediamtx --version
v1.18.3-0.20260520142345-762ae87b5f75
```

Previously:

```
go generate ./... && go install . && mediamtx --version
v1.18.2-6-111302f9
```

Ran:

```
make format lint binaries
make test
make test-32
```

Both make test and make test-32 "passed" but both printed "/bin/sh: git: not found" at the beginning. Binaries are in the form mediamtx_v1.18.2-7-gdea66680_linux_armv6.tar.gz since I have local changes. This naming comes from `git describe --tags`.

Potential follow ups:
- make binaries ask for root access. One idea would be to mount the container with the right user ID so the chown is not necessary.
- Update internal/upgrade/upgrade.go to use debug.ReadBuildInfo() so go-git is not a dependency anymore.


aler9:
Hello, the first implementation of the version retrieval system was based on `debug.ReadBuildInfo`, but it was not compatible with CI/CD at the time. Seems that this issue has disappeared though, so we can give this patch a try. Thanks for providing it.

aler9:
By reviewing this patch i noticed that `go install` is necessary for the tag to be printed correctly. Am i correct? we cannot make `go install` mandatory. Is there a way to implement this without `go install` ?

maruel:
> By reviewing this patch i noticed that `go install` is necessary for the tag to be printed correctly. Am i correct? we cannot make `go install` mandatory. Is there a way to implement this without `go install` ?

Can you clarify what you mean by go install becoming necessary?

`scripts/binaries.mk` uses `go build` as before, I haven't touched this. It is true that the git tag must be created: the binaries job in .github/workflows/release.yml is triggered exactly on git tag push so this still work by definition as before.

aler9:
As previously suspected, `info.Main.Version` is populated with the tag only when a Golang binary is compiled and installed with `go install`. In case of `go build`, which we are using for producing binaries, the tag is not available, the same applies to `go run`.

Therefore, we cannot replace the custom version retrieval system with the native one.

Issue #4485, opened 2025-05-04
Allow rpiCameraSecondary with h264 codec


InanixFR:
### Describe the feature

If I try to set the rpiCameraSecondary with a H264 codec I get this error when I start mediamtx

`ERR: supported codecs for a secondary RPI Camera stream are auto, mjpeg`

But it's possible to use several time the H264 hardware encoder with low resolution/FPS and also to use a H264 software encoder. 

I think you should allow to use H264 encoder with the rpiCameraSecondary

anacrolix:
It's not clear how to even view the stream as mjpeg

chapb:
Also interested in this feature. Alternatively, the documentation should reflect this is not possible.

aler9:
This feature is perfectly possible and i would gladly review any patch that addresses it.

chapb:
I'm happy to contribute to clarifying the documentation. 
Pardon my ignorance - I have little experience in Go and exploring the code now. Would it be a matter of changing [runSecondary](https://github.com/bluenviron/mediamtx/blob/0edc3dbe8600f6e6fba1e528080dfd754c422557/internal/staticsources/rpicamera/source.go#L283) to be more like runPrimary, or is this something that would need to be tackled in the upstream [mediamtx-rpicamera](https://github.com/bluenviron/mediamtx-rpicamera) project as well?

aler9:
You need to update simultaneously mediamtx-rpicamera and mediamtx, it's not super easy but it's also not super hard. I'm not doing it right now just because i have more important tasks. if you have a couple of days available anyone with a minimum experience and some amount of patience can do that.
I'm not updating the documentation because it's not a hard constraint, it's an implementation constraint that can be overcome.

chapb:
Sorry, to clarify my language above, I meant it is not currently supported vs not possible. I added some language in the documentation example so that others wouldn't be confused if they were following the example and presumed M-JPEG was the default but not the only supported codec at present (like me).

aler9:
fixed by https://github.com/bluenviron/mediamtx/pull/5898

PR #5792, opened 2026-05-21
docs: clarify codec availability for secondary stream

chapb:
Clarify the availability of codecs for secondary stream. At present, the H264 codecs are not available to a secondary stream. Current language about defaults here does not make this evident, though it can be inferred from documentation in complete configuration file. See #4485.


aler9:
not necessary anymore. See #5898

PR #5800, opened 2026-05-25
rtsp source: add rtspScale option to inject RFC 2326 Scale header on PLAY

Adiel-Sharabi:
## Summary

Adds a new per-path option `rtspScale` for the RTSP static source. When set, the value is sent as the `Scale` header in the PLAY request, per [RFC 2326 §12.34](https://datatracker.ietf.org/doc/html/rfc2326#section-12.34).

- Negative values play in reverse: `-1.0` = realtime reverse, `-2.0` = 2x reverse
- Values > 1 fast-forward; values 0 < x < 1 play in slow motion
- Default empty = header omitted = current behavior (backward-compatible)

## Why

I'm integrating MediaMTX as the broker between a Windows access-control system and a Nx Meta VMS server. The VMS honors `Scale: -1.0` to play archive recordings in reverse, which the customer needs for incident review (operator scrubs backwards through footage).

`gortsplib.Client.Play(*headers.Range)` doesn't expose a Scale parameter and `doPlay()` is unexported, so the cleanest way to inject the header is through the existing `OnRequest` hook on the static-source `Client`. When the request method is `PLAY` and `rtspScale` is non-empty, we set the header before the request leaves the client.

Other VMS servers (Pelco, Genetec, some Milestone deployments) follow the same RFC and would benefit too.

## What's in the change

| File | Change |
|---|---|
| `internal/conf/path.go` | New `RTSPScale string` field on `Path` (json: `rtspScale`) |
| `internal/staticsources/rtsp/source.go` | `OnRequest` hook injects `Scale` header on PLAY when `RTSPScale != ""` (10 lines) |
| `internal/staticsources/rtsp/source_test.go` | `TestScale` — verifies the header lands on the server side |
| `mediamtx.yml` | Documents the new field next to `rtspRangeStart` |

Total diff: **4 files, 107 insertions, 0 deletions**. No deprecations, no behavior change for users who don't set the field.

## Example

```yaml
paths:
  reverse-archive:
    source: rtsp://user:pass@vms-server:7001/camera-uuid?pos=2026-05-25T12:00:00.000Z
    sourceOnDemand: yes
    rtspScale: "-1.0"   # play backwards in realtime
```

## Verification

- `TestScale` passes (`go test ./internal/staticsources/rtsp/ -run TestScale`)
- Verified end-to-end against a real Nx Meta 6.1.1 server + Axis M5013 camera: server returns 200 OK to the PLAY with `Scale: -1.0` and streams the (server-side-interpreted) reverse playback.
- Backward compatibility: omitted/empty `rtspScale` produces identical RTSP traffic to current behavior.

## Open question for maintainers

Naming: I went with `rtspScale` to match the existing `rtspRange{Type,Start}` family. Happy to rename to `rtspPlayScale` or similar if you prefer.

## Test plan

- [ ] `go test ./...`
- [ ] Manual: configure a path with `rtspScale: "-1.0"` against any RTSP server that supports Scale; observe the header in the PLAY exchange (e.g. via `tcpdump -A` or the server's debug log)
- [ ] Confirm empty/omitted field produces identical traffic to current behavior

### DIFF
--- a/api/openapi.yaml
+++ b/api/openapi.yaml
@@ -782,6 +782,8 @@ components:
           $ref: "#/components/schemas/RTSPRangeType"
         rtspRangeStart:
           type: string
+        rtspScale:
+          type: string
         rtspUDPReadBufferSize:
           type: integer
           format: uint64
--- a/internal/conf/path.go
+++ b/internal/conf/path.go
@@ -257,6 +257,7 @@ type Path struct {
 	SourceAnyPortEnable    *bool          `json:"sourceAnyPortEnable,omitempty" deprecated:"true"`
 	RTSPRangeType          RTSPRangeType  `json:"rtspRangeType"`
 	RTSPRangeStart         string         `json:"rtspRangeStart"`
+	RTSPScale              string         `json:"rtspScale"`
 	RTSPUDPReadBufferSize  *uint          `json:"rtspUDPReadBufferSize,omitempty" deprecated:"true"`
 	RTSPUDPSourcePortRange []uint         `json:"rtspUDPSourcePortRange"`
 
--- a/internal/staticsources/rtsp/source.go
+++ b/internal/staticsources/rtsp/source.go
@@ -127,10 +127,6 @@ func (s *Source) Run(params defs.StaticSourceRunParams) error {
 		return err
 	}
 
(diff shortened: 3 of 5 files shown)


aler9:
merged, thanks!

PR #5813, opened 2026-05-29
feat(auth): kick clients and publishers when auth token expires

### LINKED ISSUE #4300: Kick clients when the JWT expires
### Describe the feature

Discussed in #3814 

Currently, JWT is checked during handshake only. If would be better if the JWT is periodically checked during streaming, in order to kick out clients with expired JWTs.

thejoeejoee:
Periodic check of auth token expiration for connected readers and publishers. Expired clients are disconnected.

Generic naming allows future extension to external auth re-validation or JWKS re-fetch.

Closes #4300


aler9:
This is interesting, my architectural concern is that in this implementation, the component in charge of periodically checking the JWT is the path, while i would have preferred readers to do this check by themselves. But the drawback of moving this logic into readers is that you have to write a dedicated check in each reader type.

Furthermore, a thing that's missing is that publisher's JWTs must be checked too.

thejoeejoee:
> This is interesting, my architectural concern is that in this implementation, the component in charge of periodically checking the JWT is the path, while i would have preferred readers to do this check by themselves. But the drawback of moving this logic into readers is that you have to write a dedicated check in each reader type.

Path felt like the right place since it already owns the lifecycle of both readers and publishers (add/remove/close) — avoids duplicating the ticker logic in each reader/publisher type. If you'd prefer it per-reader though, I could extract a shared helper to keep it DRY.

> Furthermore, a thing that's missing is that publisher's JWTs must be checked too.

they're already checked here too, checkAuthExpiry() kicks both expired readers and the publisher:

https://github.com/thejoeejoee/mediamtx/blob/feat-jwt-auth-continuous-check/internal/core/path.go#L989-L995

Issue #4874, opened 2025-08-13
Specifying webrtcAdditionalHosts with 192.168.x.x notation causes streams to fail


mjohnsonsu:
### Which version are you using?

1.14.0

### Which operating system are you using?

Linux amd64 Docker

### Describe how to replicate the issue

1. set environment variable MTX_WEBRTCADDITIONALHOSTS to "192.168.x.x,127.0.0.1" 
2. start the server
3. publish an rtsp stream
4. Connect to the webrtc stream
5. Get error Error: lookup 192.168.x.x on 192.168.0.2:53: no such host, retrying in some seconds
### Server logs

2025/08/13 09:58:40 INF [RTSP] [conn [::1]:54704] opened
2025/08/13 09:58:40 INF [RTSP] [session efa8c064] created by [::1]:54704
2025/08/13 09:58:40 INF [RTSP] [session efa8c064] is publishing to path 'main', 1 track (H264)
2025/08/13 09:58:42 INF [WebRTC] [session e550e733] created by 192.168.0.10:34465
2025/08/13 09:58:42 INF [WebRTC] [session e550e733] closed: lookup 192.168.x.x on 192.168.0.2:53: no such host

PR #5845, opened 2026-06-08
webrtc: don't abort the session when a webrtcAdditionalHosts entry can't be resolved

suMin77123:
### Description

Since #4866, hostnames in `webrtcAdditionalHosts` are resolved server-side (`net.LookupIP`) in `addAdditionalCandidates`. A resolution failure currently does `return err`, which aborts the **entire** WHEP/WHIP session — so one unresolvable entry takes down all WebRTC playback, even when the other entries are valid.

This breaks setups where a host legitimately can't be resolved on the server:
- air-gapped / offline networks (no DNS on the server), and
- split-horizon / overlay-network DNS where the name only resolves on the client (e.g. Tailscale MagicDNS).

(Related: #4874 — same root cause, surfaced with an invalid value and closed as not-planned.)

### Change

On resolution failure, log a warning and advertise the literal hostname as the ICE candidate (so the remote peer can resolve it) instead of failing. The resolvable path is unchanged — entries that resolve still produce IP candidates — so clients that rely on server-side resolution (and Firefox/Safari, which don't resolve FQDN candidates) are unaffected; this only changes the otherwise-fatal case.

If you'd prefer to simply skip the unresolvable host (without advertising a literal candidate), I'm happy to switch to that — the key fix is not aborting the whole session.

### Test

Added `TestPeerConnectionAdditionalHostsUnresolvable`: an unresolvable host (`*.invalid`, RFC 6761) no longer fails `CreateFullAnswer` and is advertised as a literal `typ host` candidate.

### DIFF
--- a/internal/protocols/webrtc/peer_connection.go
+++ b/internal/protocols/webrtc/peer_connection.go
@@ -504,7 +504,12 @@ func (co *PeerConnection) addAdditionalCandidates(firstMedia *sdp.MediaDescripti
 		} else {
 			tmp, err := net.LookupIP(host)
 			if err != nil {
-				return err
+				// The host can't be resolved server-side - e.g. in air-gapped
+				// networks without DNS, or with split-horizon / overlay DNS
+				// names that only resolve on the client. Skip it instead of
+				// failing the entire session, so the other entries still work.
+				co.Log.Log(logger.Warn, "cannot resolve additional host %q, skipping it: %v", host, err)
+				continue
 			}
 
 			ips = make([]string, len(tmp))
--- a/internal/protocols/webrtc/peer_connection_test.go
+++ b/internal/protocols/webrtc/peer_connection_test.go
@@ -886,3 +886,50 @@ func TestPeerConnectionPublishDataChannel(t *testing.T) {
 
 	<-dataReceived
 }
+
+func TestPeerConnectionAdditionalHostsUnresolvable(t *testing.T) {
+	// A host in AdditionalHosts that can't be resolved server-side - for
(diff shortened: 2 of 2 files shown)


internal/protocols/webrtc/peer_connection.go
```diff
@@ -504,12 +504,18 @@ func (co *PeerConnection) addAdditionalCandidates(firstMedia *sdp.MediaDescripti
...
-			ips = make([]string, len(tmp))
-			for i, e := range tmp {
-				ips[i] = e.String()
+				// The host can't be resolved server-side. This happens in
+				// air-gapped networks without DNS, or with split-horizon /
+				// overlay DNS names that only resolve on the client.
+				// Advertise the literal hostname so the remote peer can
+				// resolve it, instead of failing the entire session.
+				co.Log.Log(logger.Warn, "cannot resolve additional host %q, advertising it as-is: %v", host, err)
+				ips = []string{host}
```
aler9: we don't want host entries to reach clients. Chrome is the only browser that supports this, all others incur in an error. Remove this

aler9 (changes requested):

internal/protocols/webrtc/peer_connection.go
suMin77123: Removed — switched to skipping the unresolvable host (warn + `continue`) instead, so a single unresolvable entry no longer aborts the whole session while the other (resolvable) entries still produce candidates. No host entries reach clients now.

If you'd rather keep erroring out on an unresolvable host, I'm happy to close this instead.

aler9 (approved):

aler9:
merged, thanks!

PR #5867, opened 2026-06-11
fix(playback): avoid panic when MP4 muxer flushes with no samples

### LINKED ISSUE #5865: Playback: panic (nil pointer) in muxerMP4.flush on /get?format=mp4 when recording contains an audio track unsupported by plain MP4 — whole process exits

v1.19.0

Linux arm64 standard

### Describe the issue

A `GET /get?...&format=mp4` request to the playback server can crash the **whole MediaMTX process** with a nil-pointer panic in the plain-MP4 muxer:

```
panic: runtime error: invalid memory address or nil pointer dereference
github.com/bluenviron/mediamtx/internal/playback.(*muxerMP4).flush(...)
        /s/internal/playback/muxer_mp4.go:93
github.com/bluenviron/mediamtx/internal/playback.seekAndMux(...)
        /s/internal/playback/on_get.go:111
github.com/bluenviron/mediamtx/internal/playback.(*Server).onGet(...)
        /s/internal/playback/on_get.go:181
```

Because the HTTP handler is wrapped in `handlerExitOnPanic`, the panic terminates the entire process — every active stream and ongoing recording is dropped until systemd restarts the service. The HTTP client just sees "server disconnected without sending a response".

It happens consistently (4/4 attempts, plus later attempts on other time ranges of the same path) for a path whose **recording contains an audio track that a plain (non-fragmented) MP4 cannot carry**:

- The camera publishes H264 video + audio over RTSP; recording is `recordFormat: fmp4` (segments record fine).
- `ffprobe` on a recorded segment shows: stream #0 `h264/video`, stream #1 `codec_type=audio codec_name=unknown`.
- Remuxing that same segment with `ffmpeg -i seg.mp4 -c copy -movflags +faststart out.mp4` fails with `Could not find tag for codec none in stream #1, codec not currently supported in container` (works with `-map 0:v`), which suggests the audio codec has no representation in plain MP4.
- Other paths (cameras without that audio track) play back fine with `format=mp4` on the same setup.

So it looks like `muxerMP4` ends up with an uninitialized/nil track for the unsupported audio codec and dereferences it in `flush()`. Expected behaviour would be an HTTP 4xx/5xx (or silently skipping the unsupported track), never a process-wide crash.

1. start MediaMTX with a path published by an RTSP camera that sends H264 video + an audio codec that plain MP4 cannot mux (ffprobe reports it as `unknown` in the recorded fmp4)
2. enable recording: `record: yes`, `recordFormat: fmp4`
3. wait for at least one segment, then request:
   `GET http://<host>:9996/get?path=mycam&start=<time-inside-recording>&duration=20&format=mp4`
4. the process panics with the stack above and exits (all sessions dropped)

### Server logs

```
Jun 11 16:03:25 host mediamtx[12164]: panic: runtime error: invalid memory address or nil pointer dereference
Jun 11 16:03:25 host mediamtx[12164]: github.com/bluenviron/mediamtx/internal/protocols/httpp.(*handlerExitOnPanic).ServeHTTP.func1()
Jun 11 16:03:25 host mediamtx[12164]:         /s/internal/protocols/httpp/handler_exit_on_panic.go:21 +0x50
Jun 11 16:03:25 host mediamtx[12164]: panic({0x100dec0?, 0x2c4bf50?})
Jun 11 16:03:25 host mediamtx[12164]:         /usr/local/go/src/runtime/panic.go:860 +0x12c
Jun 11 16:03:25 host mediamtx[12164]: github.com/bluenviron/mediamtx/internal/playback.(*muxerMP4).flush(0x1d9b288?)
Jun 11 16:03:25 host mediamtx[12164]:         /s/internal/playback/muxer_mp4.go:93 +0x20
Jun 11 16:03:25 host mediamtx[12164]: github.com/bluenviron/mediamtx/internal/playback.seekAndMux({0x499e4391a5f4?, 0x499e4345e00e?}, {0x499e434b9d48, 0x2, 0x117}, {0x361?, 0x1?, 0x0?}, 0x4a817c800, {0x1da7280, ...})
Jun 11 16:03:25 host mediamtx[12164]:         /s/internal/playback/on_get.go:111 +0x458
Jun 11 16:03:25 host mediamtx[12164]: github.com/bluenviron/mediamtx/internal/playback.(*Server).onGet(0x499e433dc3c0, 0x499e433e4000)
Jun 11 16:03:25 host mediamtx[12164]:         /s/internal/playback/on_get.go:181 +0x46c
Jun 11 16:03:25 host mediamtx[12164]: github.com/bluenviron/mediamtx/internal/protocols/httpp.(*handlerExitOnPanic).ServeHTTP(0xffffa076fbf0?, {0x1da0db0?, 0x499e439a22e0?}, 0x499e43826160?)
Jun 11 16:03:25 host mediamtx[12164]:         /s/internal/protocols/httpp/handler_exit_on_panic.go:26 +0x48
```

(Happy to provide ffprobe output of the recorded segment or additional logs — cannot share the recording itself.)

### Network dumps

_No response_

Arturo2511:
## Issue

Fixes #5865.

A `GET /get?...&format=mp4` request can crash the **whole process** with a nil-pointer panic:

```
panic: runtime error: invalid memory address or nil pointer dereference
internal/playback.(*muxerMP4).flush()   internal/playback/muxer_mp4.go:93
internal/playback.seekAndMux()          internal/playback/on_get.go:111
internal/playback.(*Server).onGet()     internal/playback/on_get.go:181
```

Because the handler is wrapped in `handlerExitOnPanic`, the panic terminates the process and every active stream/recording is dropped until the service is restarted.

## Root cause

In `seekAndMux`, `m.flush()` is always called after the parts of the requested time range have been processed. `segmentFMP4MuxParts` only calls `m.setTrack()` from the `tfdt` branch, i.e. once per `traf`. When the requested range covers a segment that yields **no samples at all** (for instance an empty / init-only segment that has a `moov` but no `moof`), no `traf` is processed, `setTrack` is never called, and `muxerMP4.curTrack` stays `nil`.

`muxerMP4.flush()` then dereferences `w.curTrack` unconditionally:

```go
if len(w.curTrack.Samples) == 0 || w.curTrack.lastDTS < 0 {
```

→ nil-pointer panic.

The fragmented muxer (`muxerFMP4.flush` → `innerFlush`) does not have this problem: it iterates over `w.tracks` and never touches `curTrack`.

## Fix

Guard `curTrack` in `muxerMP4.flush()`. A `nil` `curTrack` means no sample was written for the range, which is the same "no segments" condition already handled below, so it returns `recordstore.ErrNoSegmentsFound` (HTTP 404) instead of panicking — matching the behaviour of the fragmented-MP4 path.

## Test

Added `TestMuxerMP4NoSamples`: writes an init but no samples, then flushes. It reproduces the exact panic on the current code (`muxer_mp4.go:93`) and passes with the fix, asserting `ErrNoSegmentsFound`.

Verified locally: `go build ./...`, `go vet ./internal/playback/`, `gofmt`, and `go test -run TestMuxerMP4 ./internal/playback/` all pass.

### DIFF
--- a/internal/playback/muxer_mp4.go
+++ b/internal/playback/muxer_mp4.go
@@ -90,7 +90,7 @@ func (w *muxerMP4) writeFinalDTS(dts int64) {
 }
 
 func (w *muxerMP4) flush() error {
-	if len(w.curTrack.Samples) == 0 || w.curTrack.lastDTS < 0 {
+	if w.curTrack == nil || len(w.curTrack.Samples) == 0 || w.curTrack.lastDTS < 0 {
 		return recordstore.ErrNoSegmentsFound
 	}
 
--- a/internal/playback/muxer_mp4_test.go
+++ b/internal/playback/muxer_mp4_test.go
@@ -6,6 +6,7 @@ import (
 
 	"github.com/bluenviron/mediacommon/v2/pkg/formats/fmp4"
 	mcodecs "github.com/bluenviron/mediacommon/v2/pkg/formats/mp4/codecs"
+	"github.com/bluenviron/mediamtx/internal/recordstore"
 	"github.com/bluenviron/mediamtx/internal/test"
 	"github.com/stretchr/testify/require"
 )
@@ -56,4 +57,31 @@ func TestMuxerMP4EmptyTracks(t *testing.T) {
 	require.NoError(t, err)
 
 	require.Greater(t, buf.Len(), 0)
+}
+
+// Test that flushing without any written sample returns an error instead of panicking
+func TestMuxerMP4NoSamples(t *testing.T) {
+	var buf bytes.Buffer
+
+	mux := &muxerMP4{
+		w: &buf,
+	}
+
+	init := &fmp4.Init{
+		Tracks: []*fmp4.InitTrack{
+			{
+				ID:        1,
+				TimeScale: 90000,
+				Codec: &mcodecs.H264{
+					SPS: test.FormatH264.SPS,
+					PPS: test.FormatH264.PPS,
+				},
+			},
+		},
+	}
+
+	mux.writeInit(init)
+
+	err := mux.flush()
+	require.ErrorIs(t, err, recordstore.ErrNoSegmentsFound)
 }


aler9:
merged, thanks!

PR #5873, opened 2026-06-15
fix: redact sensitive headers in HTTP debug logs

varun-nominal:
## Summary
- redact credential-bearing HTTP request headers before writing debug request dumps
- keep the original request untouched for downstream handlers
- add a regression test that verifies sensitive header values are not present in the dumped log output

## Why
HTTP debug logs currently dump incoming requests, including headers. When clients authenticate with headers such as `Authorization: Bearer ...`, enabling debug logging can expose credentials in application logs.

## Validation
- `docker run --rm -v "$PWD":/src -w /src golang:1.26-alpine3.24 sh -lc '/usr/local/go/bin/gofmt -w internal/protocols/httpp/handler_logger.go internal/protocols/httpp/handler_logger_test.go && /usr/local/go/bin/go test ./internal/protocols/httpp'`

### DIFF
--- a/internal/protocols/httpp/handler_logger.go
+++ b/internal/protocols/httpp/handler_logger.go
@@ -14,6 +14,28 @@ const (
 	maxDumpedRequestBodySize = 10 * 1024
 )
 
+var requestHeadersToRedact = map[string]struct{}{
+	"Authorization":       {},
+	"Cookie":              {},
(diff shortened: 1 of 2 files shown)


varun-nominal:
@aler9 noticed that we could be logging tokens

aler9:
merged, thanks!

PR #5912, opened 2026-06-30
feat: support forcing internal authentication for selected actions

midoelhawy:
This change adds a new configuration option, `authHTTPForceInternalUsersForActions`, to allow specific actions to be authenticated using `authInternalUsers` even when `authMethod` is set to `http`.

This is useful when most actions should be authenticated by an external HTTP authentication service, while a small subset (for example `publish` or `api`) should use the built-in user database.

Example:

```yaml
authMethod: http

authHTTPExclude:
  - action: publish
  - action: api
  - action: metrics
  - action: pprof

authHTTPForceInternalUsersForActions:
  - action: publish
  - action: api

authInternalUsers:
  - user: internal
    pass: internal_pass
    permissions:
      - action: publish
      - action: api
```

With this configuration:

* `publish` and `api` are authenticated using `authInternalUsers`.
* `metrics` and `pprof` remain excluded from HTTP authentication.
* All other actions continue to use the configured HTTP authentication endpoint.


di33y-clear44:
Thank you I'm in the middle of trying now I'm almost there

midoelhawy:
@aler9 Any thoughts on this feature? Do you think it can be included, or could it cause any compatibility issues?

PR #5919, opened 2026-07-04
Use RTSP session ID in auth access requests

### LINKED ISSUE #5908: RTSP auth callback sends connection UUID instead of session UUID

1.19.1

Linux amd64 Docker

### Problem

When MediaMTX invokes the external authentication HTTP callback for RTSP read/publish requests, the `id` field contains the **connection** UUID (`conn.uuid`), not the **session** UUID (`session.uuid`).

However, when querying the API (e.g. `GET /v3/paths/{name}`), the reader's `id` field contains the **session** UUID.

This means the ID communicated during authentication can not be used to identify or kick the resulting session via the API. This _might be_ intentional (see below at workaround) but it's at least quite surprising.

### Comparison across protocols

Every other protocol is internally consistent — the same object's UUID is used in both the auth callback and the API reader response:

| Protocol | ID in auth callback | ID in API (`APIReaderDescribe`) | Consistent? |
|----------|--------------------|---------------------------------|-------------|
| WebRTC | `session.uuid` | `session.uuid` | ✅ |
| HLS | `session.uuid` | `session.uuid` | ✅ |
| RTMP | `conn.uuid` | `conn.uuid` | ✅ |
| SRT | `conn.uuid` | `conn.uuid` | ✅ |
| **RTSP** | **`conn.uuid`** | **`session.uuid`** | ❌ |

RTMP and SRT don't have a separate session abstraction, so using `conn.uuid` for both is correct. WebRTC and HLS use `session.uuid` for both. RTSP is the only protocol where auth and API disagree — it sends the connection UUID during auth but exposes the session UUID in the API.

### Root cause

In `internal/servers/rtsp/session.go`, the `PathAccessRequest` at lines 211 and 284 passes `&c.uuid` (the connection), while `APIReaderDescribe()` at line 469 returns `s.uuid` (the session):

```go
// onAnnounce (line 211) — sends conn UUID in auth request
ID: &c.uuid,

// onSetup for readers (line 284) — sends conn UUID in auth request
ID: &c.uuid,

// APIReaderDescribe (line 469) — returns session UUID in API
ID: s.uuid.String(),
```

RTSP is unique in having both a conn and a session as distinct object and the auth code incorrectly references the connection rather than the session.

### Why this matters
External auth servers that track session IDs (received during the auth callback) to later revoke access (by kicking sessions matched from the path API) cannot function for RTSP consumers. The stored ID will never match any reader in the path listing, making targeted session termination impossible.

### Workaround

The connection object is queryable via `GET /v3/rtspconns/get/{id}` and exposes a `session` field mapping the connection UUID to its session UUID. An auth server can resolve the ID from the callback to the actual session UUID, then use it for kick/match operations. It's conceivable this indirection is by design (authenticate the connection, derive the session), but it would be surprising given every other protocol provides a directly usable ID in the auth callback.

### Suggested fix

Fixing the issue is easy: in `internal/servers/rtsp/session.go`, change the two `PathAccessRequest` structs to pass `&s.uuid` instead of `&c.uuid`:

```go
// line 211 (onAnnounce)
ID: &s.uuid,  // was &c.uuid

// line 284 (onSetup, reader path)
ID: &s.uuid,  // was &c.uuid
```

1. start MediaMTX, start an auth server that responds with 200 OK every time and logs requests: `node -e "require('http').createServer((q,s)=>{let b='';q.on('data',c=>b+=c).on('end',()=>{console.log(q.method,q.url,b);s.writeHead(200).end()})}).listen(3000)"`
2. publish with an rtsp source, e.g. `gst-launch-1.0 videotestsrc is-live=true ! videoconvert ! x264enc tune=zerolatency ! h264parse ! rtspclientsink location=rtsp://localhost:8554/rtsp-example`
3. consume the rtsp source, e.g. `ffplay rtsp://user:pass@localhost:8554/rtsp-example`
4. note down the ID in the authentication request from the auth server logs
5. try kicking the clients with `curl -X POST localhost:9997/v3/rtspsessions/kick/<id>`

### MediaMTX configuration

```yaml
# Logging
logLevel: info
logDestinations: [stdout]

# Control API: will be used to list paths and kick sessions.
api: yes
apiAddress: :9997

# Authentication: delegate all publish/read decisions to the auth server.
authMethod: http
authHTTPAddress: http://localhost:3000/auth

pathDefaults:
  record: no

paths:
  all_others:
```

### MediaMTX logs

_No response_

### Packet dump

_No response_

Cycle1337:
## Summary
- use the RTSP session UUID, instead of the connection UUID, in ANNOUNCE and SETUP auth access requests
- cover publish/read flows by asserting the auth access ID matches the API session ID

Fixes #5908

## Tests
- `go test ./internal/servers/rtsp`
- `go test ./internal/core -run RTSP` *(fails during setup because generated/embed files are missing locally: `internal/servers/hls/hls.min.js` and `internal/core/VERSION`)*


aler9:
this was OK, why did you close it? i reopened it in https://github.com/bluenviron/mediamtx/pull/5977 and gave credits.

PR #5932, opened 2026-07-07
Fix deprecated webrtcICEServers parser for IPv6 hosts; document address format

### LINKED ISSUE #5905: Fix deprecated webrtcICEServers parser for IPv6 hosts; document address format
## Problems

**1. Deprecated `webrtcICEServers` parser silently breaks for IPv6 hosts**

The parser for the deprecated `webrtcICEServers` field splits on `:` and checks for exactly 5 parts. An entry like `stun:user:pass:[2001:db8::1]:3478` contains extra colons in the host and produces the wrong part count, so the credentials are silently dropped and the URL is wrong.

**2. Address format and dual-stack behaviour are undocumented**

All address parameters accept `[host]:port`, but there is no documentation explaining what the empty-host default (`:port`) does, how to restrict to IPv4 only, or how to bind to a specific interface.

## Fixes

**1.** Replace `strings.Split(s, ":")` with `strings.SplitN(s, ":", 4)` and check `len == 4`. This keeps the hostport token intact regardless of colons in an IPv6 host.

**2.** Add an address format reference block to `apiAddress` in `mediamtx.yml` and a "Network addressing" section to the configuration docs, explaining:
- `:port` -- all interfaces; dual-stack on Linux where `bindv6only=0` (the default), IPv4 fallback otherwise
- `0.0.0.0:port` -- IPv4 only
- `[::]:port` -- explicit IPv6/dual-stack
- specific interface addresses

A pull request is attached.

poeggi:
Closes #5905

## What changes

**Bug fix** (`internal/conf/conf.go`): the deprecated `webrtcICEServers` parser used `strings.Split` and checked for exactly 5 parts. An IPv6 host contains extra colons, producing the wrong part count and silently dropping credentials. Fixed with `strings.SplitN(s, ":", 4)` -- the hostport token is kept intact regardless of colons.

**Docs** (`mediamtx.yml`, `docs/2-features/05-configuration.md`): added an address format reference explaining the `[host]:port` form, what the empty-host default does on different platforms, and how to restrict to IPv4 only or bind to a specific interface. No address default values are changed.

### DIFF
--- a/docs/2-features/05-configuration.md
+++ b/docs/2-features/05-configuration.md
@@ -49,6 +49,32 @@ There are several ways to change configuration parameters:
 
 3. Use the [Control API](22-control-api.md).
 
+## Network addressing
+
+All listener address parameters (e.g. `rtspAddress`, `rtmpAddress`, `hlsAddress`, `webrtcAddress`, etc.) use the `[host]:port` format.
+
+The default form `:port` (empty host) listens on all interfaces. On Linux, where `net.ipv6.bindv6only` defaults to `0`, this creates a dual-stack socket that accepts both IPv4 and IPv6 connections. On systems where IPv6 is disabled or `bindv6only=1`, it falls back to IPv4.
+
+To listen on IPv4 only regardless of system settings, set the host to `0.0.0.0`:
+
+```yaml
+rtspAddress: 0.0.0.0:8554
+rtmpAddress: 0.0.0.0:1935
+```
+
+To explicitly request a dual-stack socket (useful when you want to be sure IPv6 is active and not fall back silently to IPv4):
+
+```yaml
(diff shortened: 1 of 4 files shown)


poeggi:
Just rebased this onto the current main (v1.20.1) -- no conflicts, patch applies cleanly.

Is fixing the deprecated webrtcICEServers IPv6 parsing bug of interest? The issue is still present in main verbatim. Happy to keep this updated if there is a chance it gets merged, or close it if the deprecated field is on the way out entirely.

aler9:
i removed the documentation change (a dedicated PR is needed for that, furthermore i don't think the "configuration" page is fit for general tips), i added tests and merged, thanks

PR #5934, opened 2026-07-08
feat: add fallbackSource path option for seamless standby-to-live switching

poeggi:
## Summary

Adds `fallbackSource` -- a path-level option that designates a secondary static source (RTSP, RTMP, SRT, etc.) to take over seamlessly when the primary publisher is absent, then steps back down when the primary reconnects.

Includes fixes and improvements that also help the `alwaysAvailable` / `alwaysAvailableFile` live-recovery path, discovered during real-world testing:

### Fixes/improvements developed during debugging

**Gate live reactivation on first keyframe**
When a live source reconnects on an `alwaysAvailable` path, the SubStream now waits for the first IDR before becoming active. Prevents clients from receiving non-IDR video at stream start.

**Require video IDR specifically, not audio frames**
`isKeyframeUnit()` returns `true` for all audio payload types (default case). Without this fix, an audio frame arriving before the video IDR would trigger activation, unblocking non-IDR video. Fixed by checking `inMedia.Type == MediaTypeVideo` when the stream has video tracks.

**Pass live source PTS through unchanged**
On live reconnect, `ptsOffset` is set to 0 so the encoder's absolute PTS is forwarded as-is. Previously, `ptsOffset` was incorrectly set to `lastPTS + elapsed` for live sources, inflating timestamps on every reconnect. Offline fallback PTS continues to use `lastPTS + elapsed` for smooth continuation.

**RTP encoder reset on source swap**
On any source swap (either direction, either mode), the video track's RTP encoder is recreated with a new timestamp base and packetizer state. This produces a timestamp and sequence-number discontinuity that downstream decoders can use to detect the source change and resync. Scoped to H264/H265 video only (audio encoder stays continuous). Suppressed on the initial offline activation at startup -- only fires when a prior live source actually existed. Note: for RTSP readers the SSRC field is rewritten per-session by gortsplib's server layer and does not change; the observable client signal is the timestamp discontinuity.

**Docs: alternative source format requirements**
Documents that any alternative source -- whether an offline media file or a fallback stream -- ideally matches the primary stream's codec type and audio parameters; video resolution may differ (SPS/PPS are swapped automatically on source switch).

### Not in this PR

- HLS `#EXT-X-DISCONTINUITY` and MPEG-TS discontinuity indicator on source swap (possible follow-up feature)

### DIFF
--- a/api/openapi.yaml
+++ b/api/openapi.yaml
@@ -912,8 +912,7 @@ components:
           description: Fallback source
         fallbackSourceMode:
           type: string
-          enum: [preconnect, ondemand]
-          description: Fallback source connection mode
+          description: "\"preconnect\" (default) or \"ondemand\""
         maxReaders:
           type: integer
           format: int64
--- a/api/openapi.yaml
+++ b/api/openapi.yaml
@@ -907,6 +907,12 @@ components:
           type: string
           nullable: true
           deprecated: true
+        fallbackSource:
+          type: string
+          description: Fallback source
+        fallbackSourceMode:
+          type: string
+          description: "\"preconnect\" (default) or \"ondemand\""
         maxReaders:
           type: integer
           format: int64
--- a/api/openapi.yaml
+++ b/api/openapi.yaml
@@ -912,7 +912,8 @@ components:
           description: Fallback source
         fallbackSourceMode:
           type: string
-          description: "\"preconnect\" (default) or \"ondemand\""
+          enum: [preconnect, ondemand]
+          description: Fallback source connection mode
         maxReaders:
           type: integer
           format: int64
--- a/api/openapi.yaml
+++ b/api/openapi.yaml
@@ -980,7 +980,8 @@ components:
           description: Fallback source
         fallbackSourceMode:
           type: string
-          description: "\"preconnect\" (default) or \"ondemand\""
+          enum: [preconnect, ondemand]
+          description: Fallback source connection mode
         forward:
           type: array
           description: Forward
--- a/api/openapi.yaml
+++ b/api/openapi.yaml
@@ -975,6 +975,12 @@ components:
           type: string
           nullable: true
           deprecated: true
+        fallbackSource:
+          type: string
+          description: Fallback source
+        fallbackSourceMode:
+          type: string
+          description: "\"preconnect\" (default) or \"ondemand\""
         forward:
           type: array
           description: Forward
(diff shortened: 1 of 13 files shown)


poeggi:
ready for review

poeggi:
Identified some issues during testing - working on it.

poeggi:
Tested the full failover cycle end-to-end with a local H.264 RTSP publisher (cam) and an AlwaysAvailable H.264 filler path as the FallbackSource.

**Test config:**
`yaml
paths:
  filler:
    alwaysAvailable: yes
    alwaysAvailableTracks:
      - codec: H264
  cam:
    fallbackSource: rtsp://127.0.0.1:8554/filler
    fallbackSourceMode: preconnect
`

**Cycle observed (log evidence):**

1. Startup: filler available, fallback source connects and parks, cam comes online with fallback stream.
2. Primary publisher connects: activates on first keyframe, fallback steps aside.
3. Primary disconnects: fallback activates immediately (no IDR gate on primary-drop path).
4. Primary reconnects: activates on next keyframe, fallback steps back down.

All four transitions logged correctly and cleanly. No stuck state, no double-activation, no stream gap beyond one GOP on reconnect.

**RTP timestamps:** confirmed pass-through for UseRTPPackets=true sources (primary publisher). gortsplib only overwrites SSRC on the server send path -- Timestamp field is untouched. Clients signal a source swap via timestamp discontinuity from the RTP encoder reset (committed in df196d7).

Branch also rebased onto upstream main as of today (2026-07-19), incorporating the offline_sub_stream_track.go PTS fix (#5960) and the UDP port range change (#5958).

The-Padi:
This would ba a great feature. Look forward to it !

poeggi:
Rebased onto v1.21.0 and fixed the lint and formatting fallout from the rebase, so all GitHub Actions jobs are green again.

Two cleanups went in on top:

- `SubStream.activate()` holds `Stream.mutex` across `initialize2()`. `initialize2()` calls `writeUnit()` internally to push codec parameters to readers, so this is the same hazard you fixed for `Stream.Initialize()` in 550bb297.
- Removed a write-only `HasFallbackSource` field from `Stream`. It was set in two places in the path code and read nowhere, since the swap is driven by the per-SubStream `FallbackSwap` flag. Behaviour-neutral.

Still open on my side, recording it here so it is visible rather than silently missing: **unit test coverage**. codecov/patch is at 17.8% and that is the only red check left. I want to exercise the full failover cycle on real hardware first (primary active -> primary drops -> fallback takes over -> primary reconnects -> fallback steps back down) so the tests encode observed behaviour instead of my assumptions about it. I will push tests once that is done, and I am not asking for a merge before then.

If the overall approach looks wrong to you, I would rather hear that now than after writing the tests.

PR #5936, opened 2026-07-08
perf: preload alwaysAvailableFile into memory with configurable size threshold

### LINKED ISSUE #5935: perf: preload alwaysAvailableFile into memory to reduce disk I/O on loop
## Use case

`alwaysAvailableFile` currently opens the MP4 from disk on every playback loop iteration. For short files used as offline slates, this causes unnecessary disk I/O on every loop restart and adds latency on the first switch from primary to offline.

## Proposed change

Preload files up to a configurable threshold into memory. Files larger than the threshold fall back to disk access (existing behaviour). The threshold defaults to 16 MB and is configurable per path.

```yaml
paths:
  live:
    alwaysAvailableFile: /path/to/slate.mp4
    alwaysAvailablePreloadSize: 16777216  # 16 MB, 0 = always read from disk
```

poeggi:
Closes #5935

## What changes

Adds `alwaysAvailablePreloadSize` path configuration field (default 16 MB).

- Files up to the threshold are read into memory on the first `StartOfflineSubStream()` call and reused on every loop iteration; the cache is invalidated when the file mtime changes.
- Files larger than the threshold are opened from disk per loop iteration (existing behaviour).
- Set to `0` to always read from disk.

No file size limit is enforced; large files simply bypass the cache.


poeggi:
Ready for review

poeggi:
identified an issue during testing - changed to "WIP" for now.

poeggi:
Tested and works - ready for review.

aler9:
Hello, while i perfectly understand the usefulness of loading `alwaysAvailableFiles` into memory, the same exact feature is already provided by all operating systems and some hardware, in at least 3 ways:
* many disks come with a cache zone that is automatically filled with the most recent or most accessed data.
* the file system driver in the operating system already caches in memory files that are often accessed.
* operating systems already provide a way to create an in-memory temporary file system and write and read files there.

Therefore, short files are probably already cached inside the memory by the operating system, and, if they are not, they can be moved in-memory through a temporary file system, that by the way supports file sizes way larger than 16MB.

When a feature is already provided by the operating system or another software, the general approach is to avoid rewriting that feature inside the server, in order to avoid maintaining code that does not represent added value and at the same time to better integrate with the existing ecosystem.

poeggi:
I think the idea is to make it easy and reproducible for the user.

Mechanism 1 + 2 you mention (HW cache or OS Cache) only work for repeated access at a certain repetition rate. And even then not on first access. So non-deterministic and unreliable for quite some usecases.

Mechanism 3), correct, can be used, but requires special care and awareness of the user. Furthermore it would not mitigate the effects of the FS specific open, handle creation, allocation and data load times needed. They may arguably be small (on most platforms).

So while I agree, there is other similar mechanisms, I think neither is the same degree of "just works" as is this implementation proposal.

As always, YMMV.

poeggi:
Rebased onto v1.21.0 and squashed to a single commit. All CI jobs are green.

The rebase had silently dropped three things, now restored: the `alwaysAvailablePreloadSize` default in the `conf_test.go` expectation, the property in `api/openapi.yaml`, and it left an `err` shadow in `stream.go`.

Also added a `file preloaded` case to `TestStreamAlwaysAvailable`, which exercises the new path end to end: the file is read into memory during `StartOfflineSubStream()` and the offline track is then served from the byte slice instead of a file handle.

Happy to adjust the default (currently 16 MiB) or the option name if you would prefer something else.

PR #5940, opened 2026-07-08
Add wall-clock alignment for recording segments

Shulepov:
Recording segments are currently split relative to the moment recording starts. For example, with `recordSegmentDuration: 1h`, if a stream starts or reconnects at 12:13, segments are created
  around 12:13-13:13, 13:13-14:13, and so on.

  For archival recording, it is often more useful to split files on wall-clock boundaries, such as 12:00-13:00, 13:00-14:00, or 12:00-12:05, 12:05-12:10. This makes recorded files easier to
  browse, correlate with external systems, export, and manage with retention or post-processing jobs.

  This PR adds an optional `recordSegmentDurationAligned` setting. When enabled, segment rotation is aligned to wall-clock boundaries based on `recordSegmentDuration`. The default remains
  unchanged, so existing configurations keep the current behavior.

  The first segment after startup or reconnection can be shorter than `recordSegmentDuration`, and actual segment switches still happen on the first suitable random access point near the aligned
  boundary.

### DIFF
--- a/api/openapi.yaml
+++ b/api/openapi.yaml
@@ -970,6 +970,8 @@ components:
           type: string
         recordSegmentDuration:
           type: string
+        recordSegmentDurationAligned:
+          type: boolean
         rpiCameraAWB:
           type: string
         rpiCameraAWBGains:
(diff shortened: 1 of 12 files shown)


stephanecharette:
Any chance this will be accepted?  We were looking for this exact feature today.

PR #5962, opened 2026-07-19
api: add endpoints to start/stop recording on a path at runtime

### LINKED ISSUE #3720: start/stop recording by http like: `http://localhost:8889/mystream001/recording?state=start`
## Describe the feature

Description

Can we add an http interface to control the recording of the stream? To control the recording status of multiple streams.

for example: 

```bash
http://localhost:8889/mystream001/recording?state=start
http://localhost:8889/mystream002/recording?state=stop
```

or
```bash
http://localhost:9997/v3/recordings/?state=start&stream=mystream001
```

If `record:yes` is configured in `mediamtx.yml`, it seems that it will cause multiple streams to be recorded at the same time

mrasadi:
## What

Adds two Control API endpoints to start and stop recording on a path at
runtime, without touching its persisted configuration and without requiring
the publisher to reconnect:

```
POST /v3/recordings/start/{name}
POST /v3/recordings/stop/{name}
```

Closes #3720.

## Why

Today the only way to influence recording is the `record: yes/no` field in a
path's config. Toggling it at runtime means going through
`POST /v3/config/paths/patch/{name}`, which is a config mutation rather than
a dedicated session command. As reported in #3720, this gets awkward when
multiple streams share a single path with `record: yes` — patching the
shared config affects everything matching that path, not just the stream the
caller cares about.

## Design notes / deviation from the issue's proposed shape

The issue proposes a single endpoint toggled by a query parameter:

```
POST /v3/recordings/{streamName}?state=start
POST /v3/recordings/{streamName}?state=stop
```

I went with two explicit endpoints (`/recordings/start/{name}` and
`/recordings/stop/{name}`) instead, to match the existing Control API
convention of encoding the action as a path segment rather than a query
parameter (see `/config/paths/add`, `/patch`, `/replace`, `/delete`, and
`/*sessions/kick`). This keeps the new routes consistent with the rest of
`api.go` and avoids introducing a new "verb-in-query-string" pattern. Happy
to switch to the query-param shape if maintainers prefer to match the issue
literally.

## Behavior

- **Start**: fails with `400` if the path has no active publisher (nothing
  to record). If the path is already recording, the request is a no-op and
  returns `200` — this makes it safe to call from multiple places without
  worrying about interleaved/concurrent calls restarting the recorder and
  cutting a new segment unnecessarily.
- **Stop**: fails with `400` if the path is not currently recording.
- Both return `404` if the path doesn't exist.
- Neither endpoint disturbs the publisher, any attached reader, or the
  stream itself — only the recorder is affected.
- These calls do **not** persist: they act directly on the path's live
  recorder, the same internal mechanism the config-driven `record: yes`
  already uses (`path.startRecording()` / closing `path.recorder`). The
  path's `record` config value is left untouched. A config reload that
  doesn't change `record` won't undo the runtime override; a publisher
  reconnect will re-evaluate `record` from config as it does today, so a
  runtime override doesn't outlive the current publish session unless the
  config is also updated.

## Other changes

- `GET /v3/paths/get/{name}` and `GET /v3/paths/list` now include a
  `recording: bool` field reflecting whether the path currently has an
  active recorder, so callers can check state before/after calling the new
  endpoints without going through `/v3/recordings/get/{name}` (which lists
  historical segments, not live state).
- `api/openapi.yaml` regenerated from `internal/apidocsgen/openapi.template.yaml`
  via `go run ./internal/apidocsgen` (per `scripts/apidocs.mk`).

## Testing

- `internal/api`: table-driven unit tests (`TestRecordingStart`,
  `TestRecordingStop`) covering success, path-not-found, and
  not-publishing/not-recording error cases, following the existing
  `testPathManager` mock pattern.
- `internal/core`: `TestPathAPIRecordingStartStop`, a full integration test
  using a real RTSP publisher (`newInstance` + `gortsplib.Client`), that:
  - confirms starting recording on a non-publishing path returns 400
  - confirms starting on a nonexistent path returns 404
  - publishes a stream with `record: no` in config, verifies no segment is
    written
  - starts recording via the new API mid-stream (no reconnect), verifies a
    segment file appears and `recording: true` is reported via
    `/v3/paths/get`
  - confirms a second start call is idempotent (still 200)
  - stops recording via the API, verifies no further segments are written
    and `recording: false` is reported
  - confirms a second stop call returns 400
  - confirms the publisher/stream keep working after both transitions
- Existing tests (`TestPathRecord`, `TestPathManagerConfigHotReload`,
  `TestPathsList`, `TestPathsGet`, full `internal/core` and `internal/api`
  suites) still pass unmodified in behavior — only mock path managers in
  `internal/metrics` and `internal/api` gained stub implementations of the
  two new `defs.APIPathManager` interface methods.
- `go build ./...`, `go vet ./internal/...`, and the full test suites above
  were run locally against Go 1.26 (this repo's declared toolchain).

## Files touched

- `internal/defs/api_path.go` — extend `APIPathManager` interface with
  `APIRecordingStart`/`APIRecordingStop`; add `Recording` field to `APIPath`
- `internal/core/path.go` — new per-path channels/handlers
  (`doAPIRecordingStart`/`doAPIRecordingStop`) wired into the path's event
  loop, plus exported `apiRecordingStart`/`apiRecordingStop`
- `internal/core/path_manager.go` — `APIRecordingStart`/`APIRecordingStop`
  on `pathManager`, resolving the path via the existing `chAPIPathsGet`
  channel
- `internal/api/api.go` — route registration
- `internal/api/api_recordings.go` — HTTP handlers
- `internal/apidocsgen/openapi.template.yaml`, `api/openapi.yaml` — API docs
- `internal/api/api_paths_test.go`, `internal/api/api_recordings_test.go`,
  `internal/core/path_test.go`, `internal/metrics/metrics_test.go` — tests
  and mock updates

Closes #3720

### DIFF
--- a/api/openapi.yaml
+++ b/api/openapi.yaml
@@ -872,6 +872,8 @@ components:
           type: string
           nullable: true
           deprecated: true
+        recording:
+          type: boolean
         source:
           type: object
           allOf:
@@ -3746,6 +3748,87 @@ paths:
                 $ref: "#/components/schemas/Error"
         "404":
           description: connection not found.
+          content:
+            application/json:
+              schema:
+                $ref: "#/components/schemas/Error"
+        "500":
+          description: server error.
+          content:
+            application/json:
+              schema:
+                $ref: "#/components/schemas/Error"
+
+  /v3/recordings/start/{name}:
+    post:
+      operationId: recordingStart
+      tags: [Recordings]
+      summary: starts recording on a path at runtime, without altering its configuration.
+      description: >-
+        The path must have an active publisher. If the path is already
+        recording, the request succeeds without any effect.
+      parameters:
+        - name: name
+          in: path
+          required: true
+          description: name of the path.
+          schema:
+            type: string
+      responses:
+        "200":
+          description: the request was successful.
+          content:
+            application/json:
+              schema:
+                $ref: "#/components/schemas/OK"
+        "400":
+          description: invalid request, for instance the path has no active publisher.
+          content:
+            application/json:
+              schema:
+                $ref: "#/components/schemas/Error"
+        "404":
+          description: path not found.
+          content:
+            application/json:
+              schema:
+                $ref: "#/components/schemas/Error"
+        "500":
+          description: server error.
+          content:
+            application/json:
+              schema:
+                $ref: "#/components/schemas/Error"
+
+  /v3/recordings/stop/{name}:
+    post:
+      operationId: recordingStop
+      tags: [Recordings]
+      summary: stops recording on a path at runtime, without altering its configuration.
+      description: >-
+        The publisher and any reader attached to the path are left untouched.
+      parameters:
+        - name: name
+          in: path
+          required: true
+          description: name of the path.
+          schema:
+            type: string
+      responses:
+        "200":
+          description: the request was successful.
+          content:
+            application/json:
+              schema:
+                $ref: "#/components/schemas/OK"
+        "400":
+          description: invalid request, for instance the path is not recording.
+          content:
+            application/json:
+              schema:
+                $ref: "#/components/schemas/Error"
+        "404":
+          description: path not found.
           content:
             application/json:
               schema:
--- a/internal/api/api.go
+++ b/internal/api/api.go
@@ -174,6 +174,8 @@ func (a *API) Initialize() error {
 	group.GET("/recordings/list", a.onRecordingsList)
 	group.GET("/recordings/get/*name", a.onRecordingsGet)
 	group.DELETE("/recordings/deletesegment", a.onRecordingDeleteSegment)
+	group.POST("/recordings/start/*name", a.onRecordingStart)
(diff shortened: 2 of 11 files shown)


mrasadi:
I noticed #3720 was closed as not planned, and that #3718 was resolved via #4576 with a config-patch-based approach. Just flagging in case this PR is now redundant with that — happy to close it if so, or adjust if there's still value in a dedicated runtime toggle that doesn't touch persisted config. Let me know either way.

aler9:
Hello, the problem of this PR is that the same exact feature is provided by an existing technique, which is toggling the `record` parameter in the configuration of each path:

```yml
paths:
  path1:
    record: true
  path2:
    record: false
```

By using `/v3/config/paths/patch/{name}`, `record` can be toggled dynamically and the effect is the same.

If a path does not have a dedicated configuration, the server already supports adding it with `/v3/config/paths/add/{name}`.

Furthermore, both operations (toggling recording and adding configurations) were previously improved in order not to disconnect existing readers and publishers, filling every gap.

PR #6000, opened 2026-07-28
webrtc: fix packet corruption when reading G722

Lieslot:
G722 is affected by the same issue fixed for G711  in #5048.

`WriteRTPWithNTP()` overwrites the packet SSRC with a track-specific value, so the RTP header is copied first to avoid modifying packets shared across readers.


aler9:

Issue #6031, opened 2026-08-04
Recorder rejects valid H.265 streams that use temporal sub-layers (TSA_N slices): "access unit doesn't contain an IDR or non-IDR NALU


alarmasec24:
### Which version are you using?

v1.19.2 (also reproduced in isolation on v1.19.3, latest at the time of testing)

### Which operating system are you using?

Linux amd64 standard

### Describe the issue

The recorder permanently fails on H.265 RTSP sources whose encoder uses
**temporal sub-layer coding** — streams where some pictures are coded as
`TSA_N` (NAL unit type 2). The path stays `ready: true` and `bytesReceived`
keeps growing, but **nothing is ever written to disk**, and the log repeats
every ~2 seconds:

```
18:38:00 INF [path ch145] [recorder] recording 1 track (H265)
18:38:02 ERR [path ch145] [recorder] access unit doesn't contain an IDR or non-IDR NALU
18:38:04 INF [path ch145] [recorder] recording 1 track (H265)
18:38:04 ERR [path ch145] [recorder] access unit doesn't contain an IDR or non-IDR NALU
18:38:06 INF [path ch145] [recorder] recording 1 track (H265)
18:38:08 ERR [path ch145] [recorder] access unit doesn't contain an IDR or non-IDR NALU
```

The same happens with both `recordFormat: fmp4` and `recordFormat: mpegts`,
which points at the shared H.265 handling rather than either muxer.

This failure mode is dangerous in practice: because the path is `ready` and
byte counters grow, every health indicator looks green while zero footage is
recorded. We only caught it because we cross-check the segment files on disk.

## Root cause (from reading the code)

`mediacommon/pkg/codecs/h265/dts_extractor.go`, `extractInner()`: the switch
that looks for the access unit's picture slice accepts only

`IDR_W_RADL, IDR_N_LP, CRA_NUT, TRAIL_N, TRAIL_R, RASL_N, RASL_R`

and errors out otherwise. An access unit whose VCL slices are `TSA_N`
(type 2) — a perfectly legal H.265 picture type, part of temporal
scalability (Rec. ITU-T H.265 §7.4.2.2) — hits the error branch.
`TSA_R`, `STSA_N`, `STSA_R` (types 3–5) would fail the same way.

## What the failing stream looks like on the wire

We wrote a small RTSP client that groups NAL units by RTP timestamp. The
failing source (Hikvision NVR, main stream, H.265, 2688×1520, GOP 50 @ 25 fps,
[...]

Issue #5221, opened 2025-11-20
Playback does not work for DJI Drone streaming records


xarauzo:
### Which version are you using?

1.15.3

### Which operating system are you using?

Linux amd64 standard

### Describe the issue

I am using a DJI M3T drone for streaming (via Livestream feature using RTMP) to the MediaMTX server. I have "record: yes" with format "fmp4" (because I want to use the playback feature, only available on fmp4 format).

The streams are being recorded correctly, and I can find them on the corresponding path (and I can open the file with my video player). 

However, when using the "playback" functionality, the video does not play. It gets stuck on the first frame (this is using Chrome, when using Firefox it says "Video can't be played because the file is corrupt.".

On the other hand, if I stream using OBS, everything works perfectly.

### `ffprobe` result for the DJI Stream record:

```
xarauzo@LocalMachine [LOCAL] ~/Documents/mediamtx/recordings/live/1581F5FJD23AH00D1PC5 $ ffprobe 2025-11-20_16-35-17-537576.mp4 
ffprobe version 4.4.2-0ubuntu0.22.04.1 Copyright (c) 2007-2021 the FFmpeg developers
  built with gcc 11 (Ubuntu 11.2.0-19ubuntu1)
[...]

PR #6051, opened 2026-08-07
stream: merge SEI-only access units into the following picture

EricHenry:
# Summary

This PR hopes to improve / harden the remuxer for h264/h265 streams coming from sources that may be doing non expected behavior. Specifically, for this PR, when SEI access units are sent alone without any video data. 

## Background
While publishing RTMP streams from a DJI drone to MediaMTX, there was a difference in playback quality based on the streaming type that was chosen on the reading end of MediaMTX. If reading RTMP back out, the playback was as expected, however if reading WebRTC, the framerate would drop immediately. The results after investigating what could be going on lead me to find that the DJI source stream was doing some interesting things with their output RTMP stream. 

DJI's RTMP feed is packaging a single SEI access unit as one FLV tag sharing a timestamp with the the subsequent FLV tag that actually has the video data. MediaMTX's RTMP pipeline seems to [ignore these single SEI access units](https://github.com/bluenviron/mediamtx/blob/6f0faa49d189b97b51c3f83bc1aade47de0a267f/internal/protocols/rtmp/from_stream.go#L203) which resulted in the expected stream quality. 

However, it looks like the WebRTC pipeline does not ignore these and packages up the single SEI access unit into one RTP packet. This causes a doubling of packets being sent, resulting in the playback quality dropping because the frame rate is expected to be higher than the video data being sent.

## The Fix
I figured the best place to fix this would be in the unit_remuxer so that all output protocols would benefit from this change and we wouldn't need to handle it per protocol. The implementation I ended up going with is to track whether the payload is one of these single access units and hold it back in case we see an access unit with the video data next. If we do and it has the same timestamp, we merge them together. This keeps all existing streams working the same as expected, while handling this case if it comes up.

## Notes
I also included the same fix for HEVC/H265. Let me know if you want to scope it down to just h264.


ingalls:
Hey @aler9 as always really appreciate the work you do on the project! We've been running this code in production without any issues for the past 3 weeks. Is there a chance you would have a second to review this approach to see if it is something you would be willing to merge?

aler9:
Hello, there are some things that are not clear in both the explanation and the code.

> DJI's RTMP feed is packaging a single SEI access unit as one FLV tag sharing a timestamp with the the subsequent FLV tag that actually has the video data. MediaMTX's RTMP pipeline seems to [ignore these single SEI access units](https://github.com/bluenviron/mediamtx/blob/6f0faa49d189b97b51c3f83bc1aade47de0a267f/internal/protocols/rtmp/from_stream.go#L203) which resulted in the expected stream quality.

The outbound  RTMP pipeline does not ignore single SEI access units, it waits for an IDR access unit before allowing everything, as shown in the linked line of code. Thus any duplication or wrong sequence, if any, would end up in RTMP too.

> However, it looks like the WebRTC pipeline does not ignore these and packages up the single SEI access unit into one RTP packet. This causes a doubling of packets being sent, resulting in the playback quality dropping because the frame rate is expected to be higher than the video data being sent.

The outbound WebRTC pipeline is sending out exactly what it receives, there's no duplication. The only remuxing activity consists in appending VPS/SPS/PPS before every IDR. If instead the intended meaning is "when flushing out original packets, the WebRTC client behaves strangely and starts dropping packets", that's another question, and in that case i would have provided the WebRTC client name (Chrome?), version and operating system on which the problem appeared, and finally i would have produced a WebRTC-only fix rather than spamming all protocols.

EricHenry:
Hey @aler9, thanks for taking the time to review this. Apologies for the inaccuracies in my description.

For the RTMP code reference you are right on what the code was doing, I was referencing the effect that waiting for the IDR results in.

On the WebRTC side, you are right, there isn't duplication, it sends out what it receives. Again, I was inaccurately referencing the downstream effects rather than what mediaMTX is doing explicitly. The WebRTC clients (Firefox, Chrome, and Safari. I didn't get the version numbers in my testing, but it was the lastest version of each as of last month) were dropping these packets with SEI only access units.

One last apology! My intention was not to spam all protocols, but instead provide a minimal change set that could help prevent this issue for all protocols without duplicating effort / work. I'll update the PR to only target the WebRTC code path and remove it from the remuxing flow.

Thanks again for looking it over.

schimschewski:
I'm seeing what might be a related issue with a Tapo camera using H.265. see https://github.com/bluenviron/mediamtx/issues/6031#issuecomment-5214907679

I tested with MediaMTX v1.21.0, hoping that the fix from #6031 would solve it, but unfortunately I still get:

access unit doesn't contain an IDR or non-IDR NALU

I checked the stream with FFmpeg trace_headers and noticed that the camera regularly sends separate NAL type 39 (PREFIX_SEI) packets without a video NAL. They often have the same PTS/DTS as the preceding NAL type 1 video packet.

Example:
Packet: pts 11250, dts 11250
nal_unit_type = 1

Packet: pts 11250, dts 11250
nal_unit_type = 39

Could this be related to the SEI-only access units discussed here?

I can provide the complete FFmpeg trace if helpful.

EricHenry:
@schimschewski it may be, that was the behavior that I was seeing from the DJI camera streams. two packets with the same pts one of them being an SEI only access unit without any video.

Can you try my branch and copy the H.264 fix in the h.265 branch of code? It may do the trick for you.

EricHenry:
@aler9 I simplified the PR down, targeting just he H.264 WebRTC code flow. Let me know what you think, I based the work on the RTMP code path.

aler9:
Current PR does a different thing than the one initially advertised, it completely drops any AU that does not contain a IDR or a non-IDR, that is not acceptable.

Can you just share a network dump (generated with `dumpPackets: true` in the configuration) of the inbound RTMP stream from the drone that allows to replicate the issue? in this way i can take a look.

EricHenry:
@aler9 apologies for the confusion over all of this. I can work on getting you those log. The example that @schimschewski provided is the same thing that I was seeing. Two tags, same pts and dts, one having an access unit with only an SEI nal unit, and the following tag having an access unit with a vcl slice. 

What do you think would be an appropriate fix, ignoring access units with only SEI nal units in them, merging the two access units with the same time stamp together, or something else?

aler9:
In my opinion an appropriate fix would be one centered around the real compatibility issue, that first must be identified with a sufficient level of prevision. For the moment, you've discovered that access units containing a single SEI NALU, when fed into WebRTC, cause glitches with some browser (Chrome?). This is too broad, SEI NALUs can be of several kinds. Try to find out the exact SEI type that is causing the glitch, also providing a network dump as requested. This would allow to trace back the issue in Chrome source code and to find the deterministic reason why the glitch happens, not relying on empirical hypothesis.

aler9:
Thanks to the packet dump provided by @xarauzo in https://github.com/bluenviron/mediamtx/issues/5221 we now have a clear understanding of what's really happening with DJI drones.

There are two different issues:
* some drones send SEI NALUs, containing metadata, in separate access units with respect to the one containing video metadata, while they should be all in the same access unit. This is fixed at the RTMP level by https://github.com/bluenviron/gortmplib/pull/130
* These SEI NALUs, regardless of being merged with video or being standalone, are not liked by WebRTC decoders, and need to be filtered out.

aler9:
In https://github.com/bluenviron/mediamtx/pull/6243 i implemented a focused fix that filters out these problematic SEIs. The fix targets a specific SEI UUID that might not be the only one causing troubles. If the patch does not solve compatibility with your specific drone, provide a packet dump and i'll improve the fix to filter out additional UUIDs.

Issue #6059, opened 2026-08-08
Increase buffer size in MPEG-TS pre demux


benjamin-knight-sai:
### Describe the feature

For context, I have been trying to consume a live UDP multicast stream of MPEG-TS H264 encoded video and output it as a WebRTC stream. I can see in wireshark the packets come in with a size of 1358 bytes:
<img width="1043" height="140" alt="Image" src="https://github.com/user-attachments/assets/a7ee4e09-204b-4725-84b3-ef6589cdcc92" />

In the pre demuxer https://github.com/bluenviron/mediacommon/blob/main/pkg/formats/mpegts/pre_demuxer.go#L26 there is a limit of 1316 bytes for the buffer which is only 7 packets worth. If the source has larger sizes coming in, the bytes get truncated causing packets to become malformed and I assumed later dropped. 

This messes up MediaMTX's consumption of the input source and basically makes the output streams unusable. MediaMTX will accept the WebRTC connections and attempt to serve it data, but the WebRTC client will never receive any bytes. You can also see in the logs it it continually having errors parsing the KLV data stream, however this was a bit of red herring.
```shell
2026/08/07 13:51:57 WAR [path 0-eo] [MPEG-TS source] 58 decode errors, last was: unexpected sequence number: 80, expected 0                               
2026/08/07 13:51:58 WAR [path 0-eo] [MPEG-TS source] 62 decode errors, last was: unexpected sequence number: 111, expected 0              
2026/08/07 13:51:59 WAR [path 0-eo] [MPEG-TS source] 58 decode errors, last was: unexpected sequence number: 140, expected 0                              
2026/08/07 13:52:00 WAR [path 0-eo] [MPEG-TS source] 60 decode errors, last was: unexpected sequence number: 170, expected 0 
``` 

Is there any specific reason for the 7-packet specification in the `pre_demuxer`? 

I did some quick A/B testing on my own. I pulled down `mediamtx` and `mediacommon`, recreated my problem using fresh builds then increased the buffer from 1316 bytes to 1880 bytes (10 packet limit) which fixed my issue perfectly. 

I was using the following FFmpeg commands to mimic my livestream.

Known failing FFMPEG command:
```
ffmpeg -re -stream_loop -1 -i assets/video.ts -c copy -f mpegts "udp://239.1.1.1:11841?pkt_size=1354"
```
Known working FFMPEG command:
```
[...]

PR #6084, opened 2026-08-15
Support MPEG-TS streams using DVB pids - Updates go-astits from v1.15.0 to v1.16.0

ben8622:
v1.16.0 gives PMT-declared elementary streams precedence over DVB PID heuristics. This allows valid private/non-DVB MPEG-TS streams that carry H.264 on a low PID such as 0x0011 to be parsed as PES instead of being incorrectly parsed as PSI.

This fixes the MPEG-TS/WebRTC failure reproduced in bluenviron/mediamtx#6059.


aler9:
https://github.com/bluenviron/mediacommon/pull/352#issuecomment-5307183827

PR #6125, opened 2026-08-21
Register RTX for H264 incoming video tracks

### LINKED ISSUE #5678: Enable RTX for WebRTC incoming tracks
### Describe the feature

When mediamtx negotiates a WebRTC session, the media engine registers [only the primary video codec](https://github.com/bluenviron/mediamtx/blob/cae9920f00458a2a24c796144729d2628ce6af64/internal/protocols/webrtc/incoming_track.go#L30) and never registers a matching video/rtx codec.

**Expected behavior**
Each video track negotiated by mediamtx should have an associated RTX payload type in the SDP, so that NACK-driven retransmission can round-trip between mediamtx and the remote peer.

**Actual behaviour**

SDP produced by mediamtx contains only the primary video payload type; no `a=rtpmap:<pt> rtx/90000 / a=fmtp:<pt> apt=<primary>` lines are present.

jonatrios:
Fixes #5678.

`webrtc.ConfigureNack()` in `peer_connection.go` only registers the NACK RTCP interceptor. pion also needs an RTX codec explicitly registered per media codec before it can answer a publisher's retransmission offer with one -- `incomingVideoCodecs` never had one, so incoming video has no RTX payload type in the SDP at all, and NACK-triggered retransmission can't happen regardless of what the publisher offers.

Confirmed against a real GStreamer/`webrtcbin` WHIP publisher (`do-nack`/`do-retransmission` already on by default on that side): its own negotiation log went from `have 0 rtx payloads` to `have 1 rtx payloads` / `setting rtx mapping: 97 -> 99` with this change, nothing else touched. Was chasing a real WHIP publish freeze over a lossy WiFi link and traced it back to this -- retransmission being silently a no-op regardless of publisher config.

Added `TestPeerConnectionReadRegistersRTXForH264`, checked both directions: fails against the pre-fix `incomingVideoCodecs` (no RTX line in the answer at all, matching the live "have 0 rtx payloads" symptom), passes with it.

Scoped to the two H264 entries rather than the whole codec list -- the 96-127 payload type range is already at 23/32 slots between video and audio, so covering AV1/VP9/VP8/H265 too runs into that pretty fast. Renumbering the existing static assignments to make room felt like a separate, bigger decision than this fix, so left it alone rather than guess at it.

### DIFF
--- a/internal/protocols/webrtc/inbound_track.go
+++ b/internal/protocols/webrtc/inbound_track.go
@@ -106,17 +106,90 @@ var incomingVideoCodecs = []webrtc.RTPCodecParameters{
 		},
 		PayloadType: 106,
 	},
-	// RTX (RFC 4588) companions for the two H264 entries above. webrtc.ConfigureNack()
+	// RTX (RFC 4588) companions for every video codec above. webrtc.ConfigureNack()
 	// (peer_connection.go) only registers the NACK RTCP interceptor -- pion still
 	// needs an explicit RTX codec registered per media codec before it can answer a
 	// publisher's retransmission offer with one, so without this, incoming video has
 	// no RTX payload type at all and NACK-triggered retransmission can never actually
-	// happen, regardless of what the publisher offers. Scoped to H264 here (the two
-	// codecs we could validate against a real WHIP publisher) rather than every codec
-	// in this list -- the 96-127 payload type range is already at 23/32 slots used
-	// between video and audio, so covering the rest needs either renumbering some of
-	// the existing static assignments or a different scheme, which felt like a
-	// separate, bigger change from this one.
+	// happen, regardless of what the publisher offers.
+	//
+	// No renumbering needed: RegisterCodec (peer_connection.go) stores video and
+	// audio codecs in separate lists (m.videoCodecs / m.audioCodecs), so a payload
+	// type only has to be unique within this list, not across the whole 96-127
+	// range shared with incomingAudioCodecs -- confirmed against pion's own
+	// MediaEngine.RegisterCodec. 109-117 were simply unused within this list.
+	{
+		RTPCodecCapability: webrtc.RTPCodecCapability{
+			MimeType:    webrtc.MimeTypeRTX,
+			ClockRate:   90000,
+			SDPFmtpLine: "apt=96",
+		},
+		PayloadType: 109,
+	},
+	{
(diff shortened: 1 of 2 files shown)


aler9:
Hello, this is OK, but can you expand it to cover every supported codec? thanks

aler9:
i mean, at least for the video codecs. Feel free to renumber payload types.

jonatrios:
Done — RTX is now registered for every codec in `incomingVideoCodecs` (AV1, VP9's four profile variants, VP8, H265), not just the two H264 entries.

Turned out no renumbering was needed: `MediaEngine.RegisterCodec` keeps separate internal lists per media kind (`m.videoCodecs` / `m.audioCodecs`), so a payload type only needs to be unique within the video list itself, not against `incomingAudioCodecs` too. That freed up more room than I'd assumed in my first pass — 109-117 were simply sitting unused.

Test is now a table covering one payload type per codec family instead of just H264. Confirmed each case actually fails without its corresponding entry (reverted the source change locally and reran) before trusting it as a real regression test.

aler9:
Using the same payload type for a video codec and an audio codec causes known compatibility issues (https://github.com/bluenviron/mediamtx/pull/4394) (https://github.com/bluenviron/mediamtx/pull/4456). Edit the mapping in order to avoid overlapping between video and audio.

jonatrios:
You're right, thanks for the pointers. I checked pion's `MediaEngine.RegisterCodec` and it does keep separate video/audio codec lists internally, which is why I assumed the payload type only needed to be unique per media type. But that's just pion's internal bookkeeping — it doesn't reflect what Chrome actually validates, which is uniqueness across the whole BUNDLE group in the negotiated SDP. That's exactly the bug #4394 and #4456 fixed before, so I should have looked at those first.

Pushed a fix: the RTX entries that collided with audio (apt=98 through apt=104) now use 123-127 (the remaining free slots in 96-127) and 35/36 from the 30-63 range, which is the same range `reader.js` already relies on for the same reason (citing the same Chromium source). Nothing in incomingAudioCodecs or the rest of incomingVideoCodecs overlaps anymore.

Also updated the test table so the asserted RTX payload types match, and reran the full webrtc protocols package suite locally, all passing.

aler9:
i removed gaps between payload types and merged, thanks

PR #6126, opened 2026-08-21
httpp: propagate Flush() through the response writer wrappers

hcavazin:
### Problem

`handlerLogger`'s `responseRecorder` and `handlerWriteTimeout`'s `writeTimeoutWriter` both wrap `http.ResponseWriter` but implement neither `http.Flusher` nor `Unwrap()`.

Since both sit between `http.Server` and the handler, any handler that flushes mid-response has that flush silently dropped — the output only reaches the client when the handler returns.

This defeats the stated purpose of `writeTimeoutWriter`, whose own comment says it exists so one can:

> write long responses, splitted in chunks, without causing timeouts

The deadline is indeed reset per `Write()`, but nothing ever leaves the buffer until the end.

### How I ran into it

Adding a `text/event-stream` endpoint on a local branch: events written and flushed at t=0.4s only reached the client at t=2.2s, when the handler returned. The server logged each flush at the right time, which made it look like a client problem for a while.

### Fix

`Flush()` on both wrappers, plus `Unwrap()` so `http.ResponseController` can reach the underlying writer.

### Test

`TestResponseWriterWrappersPropagateFlush` drives a real listener through both wrappers and reads the body **while the handler is still blocked**, so anything that arrives arrived because of the flush and not because the response closed. It fails on current `main` by hitting the read deadline.

I kept this PR to the `httpp` fix alone — it is independent of what I was building when I found it.

### DIFF
--- a/internal/protocols/httpp/handler_flush_test.go
+++ b/internal/protocols/httpp/handler_flush_test.go
@@ -1,0 +1,90 @@
+package httpp
+
+import (
+	"bufio"
+	"bytes"
+	"context"
+	"net"
+	"net/http"
+	"testing"
+	"time"
+
+	"github.com/stretchr/testify/require"
+
+	"github.com/bluenviron/mediamtx/internal/test"
+)
+
+// A handler that flushes mid-response must have its bytes reach the client
+// before it returns. Both wrappers between the http.Server and the handler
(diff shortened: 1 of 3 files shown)


aler9:
merged, thanks

PR #6135, opened 2026-08-24
docs: Twitch Enhanced Broadcasting clarification

shevernitskiy:
Add a clarification to the Twitch forward documentation regarding Enhanced Broadcasting (HEVC and multitrack video).

Static RTMP/RTMPS forwarding only works for classic single-track H.264 streams. Twitch Enhanced Broadcasting requires an ASC (Automatic Stream Configuration) handshake and dynamic ingest endpoints, which are not supported via the standard `forward` directive.

### DIFF
--- a/docs/2-features/11-forward.md
+++ b/docs/2-features/11-forward.md
@@ -127,3 +127,8 @@ paths:
       # Also replace rtmp:// with rtmps:// in order to enable encryption in-transit.
       - dest: rtmps://ingest.global-contribute.live-video.net/app#streamKey
 ```
+
+**Warning**: Enhanced Broadcasting (2K / HEVC / Dual Format).
+Although MediaMTX supports Enhanced RTMP (including H.265/HEVC and multitrack video), the simple `forward` mechanism above only works with classic single-track H.264.
+Twitch Enhanced Broadcasting requires a special Automatic Stream Configuration (ASC) handshake that returns a dynamic ERTMP ingest endpoint. The static URL listed in the Recommended Ingest Endpoints page does not support this flow and will reject HEVC / multitrack streams.
+For Enhanced Broadcasting use a compatible client (OBS Studio with Enhanced Broadcasting enabled) that performs the ASC handshake itself.


aler9:
Hello, i'd rather add support for the ASC handshake rather than documenting that we do not support it. Do you have any documentation about it?

shevernitskiy:
> Hello, i'd rather add support for the ASC handshake rather than documenting that we do not support it. Do you have any documentation about it?

This is great! It would be very helpful for everyone using MediaMTX for restreaming to Twitch.

However, it is important to understand that this kind of handshake is not part of any specification. It is simply a service-level feature for Twitch and Amazon IVS.

Twitch itself does not provide any technical documentation at all, only a user guide explaining how to use the built-in, hardcoded mechanism in OBS. Amazon IVS uses the same handshake as well, although with slightly different parameters, and they do provide a brief description of it: https://docs.aws.amazon.com/ivs/latest/LowLatencyUserGuide/multitrack-video-sw-integration.html

One important detail is that the configuration URL is not hardcoded deep in the binary. It is stored in the services.json file that ships with OBS (and can be overridden by a user-local copy). This means a user can change the default configuration URL by editing services.json (or by preparing a separate profile that contains a modified version of it). Additionally, when using the Custom RTMP service, OBS allows providing a full multitrack configuration via the Config Override option in the stream settings, which can be used instead of fetching the configuration from a remote URL.

In short, the algorithm works roughly like this:

1. Before sending the stream, a POST request is made to the configuration URL. It contains JSON with information about the user's computer, such as the GPU, CPU, supported codecs, etc. It also contains the stream key as an authenticator, as far as I understand.
2. If the user is allowed to stream via ERTMP, they receive a JSON response containing the encoder settings for the tracks, which are automatically applied in OBS, as well as the actual RTMP URL with a generated key.
3. The streaming process then begins.

While researching this topic, I came across some very useful sources:

1. https://github.com/Windows81/Universal-OBS-Enhanced-Broadcast - this repository breaks down the process in detail.
2. https://github.com/Soulhackzlol/InstantClone - this also reverse-engineers the process to some extent.
3. It was also very useful to ask LLMs to inspect the OBS repository and explain in simple terms how this is implemented there.

I was thinking about the best way to implement this in the context of MediaMTX. I came up with a couple of ideas:

1. Proxy the configuration POST requests, replacing the target RTMP endpoint in responses (to MediaMTX) and add a new temp path with recieved dynamic rtmp url. This way, the correct multitrack configuration would be applied in OBS, the stream would be sent to MediaMTX, and MediaMTX would know where to forward it next, using the dynamically generated RTMP URL.
2. Have an option to place a static JSON file on the server and trigger a hook when the configuration is requested, leaving the orchestration to the user.

The first approach is much more user-friendly specifically for Twitch and Amazon IVS users. However, in theory it requires more work and makes MediaMTX handle things that are somewhat outside of its core scope.

Initially, I was thinking about creating my own small proxy that would dynamically create a path through the Control API, but if this feature could be built directly into MediaMTX, that would be amazing!

shevernitskiy:
It would also be useful to support a scenario where the handshake is performed for multiple specified stream keys at once. It’s a fairly common use case when you need to restream a single source to two channels simultaneously.

shevernitskiy:
While digging into how to retrieve the multitrack configuration, I discovered the following.

In response to our POST request, Twitch returns a JSON containing the OBS settings, RTMP URLs, and a temporary, dynamically generated stream key with an interesting structure. This particular key caught my attention. It can look something like this:

`v1_df098827616ccd30ff5487ecf1a74406b81e1d5eaec4df38627f1ced42b66b27_6a8e8eec_7b2276223a312c2262223a363636302c2274223a5b7b2277223a3634302c2268223a3336302c2262223a3530302c226330223a317d2c7b2277223a313932302c2268223a313038302c2262223a363030302c226330223a317d5d2c2261223a5b7b2262223a3136307d5d7d_live_50928037_q4FYJfEdu1w5yCb7as5eQZBctTGuS0S`

As you can see, it is split into several parts:

1. `v1` — obviously, the version.
2. `32 bytes` — most likely a SHA-256/HMAC signature.
3. `4 bytes` — unix timestamp hex.
4. A JSON object containing a compact representation of the track settings. It looks roughly like this:

`{"v":1,"b":6660,"t":[{"w":640,"h":360,"b":500,"c0":1},{"w":1920,"h":1080,"b":6000,"c0":1}],"a":[{"b":160}]}`

This JSON is most likely used by the ingest server to validate the incoming stream against the settings issued by the server for this particular key. Since the JSON is protected by the signature mentioned in point 2, it cannot be tampered with, making it impossible to substitute unauthorized stream parameters.

Essentially, this provides **stateless validation of the allowed stream parameters on the ingest server**.

5. The user's original stream key from the Twitch dashboard.

PR #6136, opened 2026-08-25
metrics: add remoteAddr field to forward section entry

### LINKED ISSUE #6058: Add `remoteAddr` to forward section metrics
### Describe the feature

Is it possible to add `remoteAddr="[remoteAddr]"` to forward section metrics?
It will help to correlate raw network stats (from `ss` for example) with mediamtx's pathes.
And also it will be great if we will see forwards in `/v3/paths/list`, because atm there are no readers there even with  active forwards for specific path.
Thanks!

shevernitskiy:
Add `remoteAddr` to forward destination metrics

### Motivation

There is currently no way to correlate MediaMTX outgoing forward connections with OS-level network statistics
(netstat, ss, firewall counters, etc.), because neither the metrics nor the API expose the address of the remote peer each forward destination is connected to.

This adds a remoteAddr field to the forward destinations item and a remoteAddr label to the forward_dests metric.

### Implementation

- Dest interface (internal/forward/dest.go) gains a RemoteAddr() string method; DestHandler exposes it through the existing activeDest mechanism, so it's only populated while a connection is actually active.
- Each protocol reports its peer address once the connection is established:
  - RTMP / RTSP: taken from the underlying control connection (client.NetConn().RemoteAddr()).
  - SRT: taken from conn.RemoteAddr() after dial.
  - WHIP: resolved on demand from the selected ICE candidate pair (GetSelectedCandidatePair()); new PeerConnection.RemoteAddr() helper in internal/protocols/webrtc.
- The value is empty ("") when no connection is established - i.e. while the destination is idle, during connection attempts, and in the error state between retries. This is consistent across all protocols.

### Notes

- For RTSP destinations using UDP transport, the address reported is that of the TCP control connection (server host), not the UDP flow.

Closes https://github.com/bluenviron/mediamtx/issues/6058

### DIFF
--- a/docs/2-features/22-metrics.md
+++ b/docs/2-features/22-metrics.md
@@ -156,8 +156,8 @@ moq_sessions_inbound_bytes{id="[id]",path="[path]",remoteAddr="[remoteAddr]",sta
 moq_sessions_outbound_bytes{id="[id]",path="[path]",remoteAddr="[remoteAddr]",state="[state]"} 187
 
 # Forward destinations
-forward_dests{id="[id]",path="[path]",protocol="[protocol]",state="[state]"} 1
-forward_dests_outbound_bytes{id="[id]",path="[path]",protocol="[protocol]",state="[state]"} 1234
+forward_dests{id="[id]",path="[path]",protocol="[protocol]",remoteAddr="[remoteAddr]",state="[state]"} 1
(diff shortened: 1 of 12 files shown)


aler9:
Hello, thanks for providing this PR. I took the occasion to review the API and metrics interface, and i thought that it would have been better to introduce a new system for exposing multiple protocol-specific statistics, rather than using the existing one to add a single, generic statistic (`remoteAddr`).

Therefore, in #6178 i added a new `typeSpecific` property to forward dests that can contain several protocol-specific statistics, including `remoteAddr`. Thanks for providing this patch though.

PR #6162, opened 2026-08-31
confwatcher: add polling fallback for changes fsnotify doesn't report

### LINKED ISSUE #2661: Hot reloading not working

v1.2.1

- [ ] Linux amd64 standard
- [x] Linux amd64 Docker
- [ ] Linux arm64 standard
- [ ] Linux arm64 Docker
- [ ] Linux arm7 standard
- [ ] Linux arm7 Docker
- [ ] Linux arm6 standard
- [ ] Linux arm6 Docker
- [ ] Windows amd64 standard
- [ ] Windows amd64 Docker (WSL backend)
- [ ] macOS amd64 standard
- [ ] macOS amd64 Docker
- [ ] Other (please describe)

## Describe the issue

When changing the configuration file nothing is happening, nor logs nor any change on the server.

1. start the server - docker run --rm -it --network=host -v $PWD/mediamtx.yml:/mediamtx.yml bluenviron/mediamtx  
2.update config file
 
## Did you attach the server logs?

no

It does not shown any logs even in debug mode when the file is updated manually

## Did you attach a network dump?

no

fra-shipper:
## What

Adds a periodic stat-based polling backstop to `ConfWatcher` so hot-reload still works when fsnotify never receives events for the watched configuration file, as reported in #2661 for Docker bind mounts.

## Why

`ConfWatcher` relies solely on fsnotify watching the parent directory of the configuration file. On some setups — most commonly a configuration file exposed into a container through a Docker bind mount — changes made outside the container never generate an fsnotify event inside it, because fsnotify only watches the parent directory and Docker's bind-mount propagation does not forward host-side inotify events into the container. As a result, hot reloading silently does nothing: no logs, no error, no reload.

This has been reported repeatedly since 2023 (#2661, reopened, with maintainer confirmation that it's a real Docker limitation), and there is no existing fallback path in the code.

## Fix

`ConfWatcher.run()` now also runs a `time.Ticker`-based poll (every second) alongside the existing fsnotify watch. On each tick it `os.Stat()`s the watched file and compares its modification time and size (via a small `fileState`/`equal()` helper, to avoid a direct `time.Time` `!=` comparison) against the last known values; if they differ, it waits `additionalWait` and re-stats — mirroring the existing fsnotify branch's "let the writer finish" behavior — before delivering a signal through the same channel and `minInterval` dedup path used by the fsnotify branch. fsnotify remains the fast path for setups where it works correctly; polling is purely a backstop, and the two paths share `lastState`/`lastCalled` so a change reported by one is not re-reported by the other.

While adding the poll path, a pre-existing edge case surfaced: on a throttled fsnotify event (two writes within `minInterval`), the code used to skip updating `lastState` entirely, leaving it pointing at the first, already-signaled write. Once `minInterval` elapsed, the poll ticker would then compare the real (second-write) mtime against that stale `lastState`, see a mismatch, and fire a spurious extra reload roughly one poll interval later. `lastState` is now refreshed whenever a real on-disk change is observed, whether or not the signal itself gets throttled, so the poll ticker no longer disagrees with what fsnotify already handled.

The hot-reloading paragraph in `docs/2-features/05-configuration.md` now mentions the polling fallback.

## Testing

All verification was run inside `golang:1.26-alpine` via a local Linux VM (Colima/Docker), matching this repo's CI (ubuntu-24.04, Go 1.26, golangci-lint v2.13.2, `RACE=-race`). macOS was intentionally not used for the authoritative runs, since fsnotify's kqueue backend follows symlinks differently than Linux inotify and would give a false pass on the new test.

- Added `TestPollFallback`, which reproduces the "fsnotify never fires" condition without needing Docker: the watched path is a symlink whose target lives in a different, unwatched directory, so fsnotify (which only watches the symlink's parent directory) never sees an event for writes to the target, exercising the poll path exclusively.
- RED (pre-fix `confwatcher.go` + new test): `go test ./internal/confwatcher/... -v` → `confwatcher_test.go:183: timed out`, `--- FAIL: TestPollFallback (3.02s)`, `FAIL github.com/bluenviron/mediamtx/internal/confwatcher [exit FAIL]`; the 5 pre-existing tests still passed unmodified.
- GREEN (fix applied): `go test ./internal/confwatcher/... -v -count=1` → `--- PASS: TestNoFile (0.00s)`, `--- PASS: TestWrite (0.01s)`, `--- PASS: TestWriteMultipleTimes (0.52s)`, `--- PASS: TestDeleteCreate (0.03s)`, `--- PASS: TestSymlinkDeleteCreate (0.02s)`, `--- PASS: TestPollFallback (1.01s)`, `PASS`, `ok github.com/bluenviron/mediamtx/internal/confwatcher 1.587s`.
- A follow-up review pass caught the throttled-event/spurious-reload edge case described above. `TestWriteMultipleTimes`'s post-signal wait window was extended from 500ms to 2500ms (longer than `minInterval` + `pollInterval`) so it also exercises the poll ticker; this reproduces the regression against the pre-fix code (`TestWriteMultipleTimes` fails at ~2.01s with "should not happen") and passes once the `lastState` refresh fix is applied (full 2.52s, genuinely exercising the poll-tick window).
- Race/flake check: `go test ./internal/confwatcher/... -race -count=3 -v` (`CGO_ENABLED=1`) → all 18 sub-runs (6 tests × 3 reps) pass, `ok github.com/bluenviron/mediamtx/internal/confwatcher 5.765s`, no data races reported.
- Lint: `golangci-lint run ./internal/confwatcher/...` (v2.13.2, 30 active linters incl. gofmt/gofumpt/govet/staticcheck/gocritic/revive/dupl/errcheck) → `0 issues`.
- `gofmt -l internal/confwatcher/` → clean, on both macOS host and inside the Linux container.
- `go vet` and `go build` on `internal/confwatcher` and `internal/certloader` (the other `ConfWatcher` consumer, cert/key hot-reload) → clean on macOS host, used only to confirm the package compiles standalone.
- Not run: the full `make test`/`make test-internal` suite (whole-repo, well beyond the scope of this 2-file change) and `make lint-conf`/`lint-docslinks`/`lint-docsorder` (no docs schema or link structure touched — only prose was added to an existing paragraph).

Diff is surgical: `internal/confwatcher/confwatcher.go`, `internal/confwatcher/confwatcher_test.go`, and one sentence added to `docs/2-features/05-configuration.md`. No unrelated refactors, no new dependencies.

Fixes #2661

### DIFF
--- a/docs/2-features/05-configuration.md
+++ b/docs/2-features/05-configuration.md
@@ -14,7 +14,7 @@ There are several ways to change configuration parameters:
      docker run --rm -it --network=host -v "$PWD/mediamtx.yml:/mediamtx.yml:ro" bluenviron/mediamtx:1
      ```
 
-   The configuration can be changed dynamically when the server is running (hot reloading) by writing to the configuration file. Changes are detected and applied without disconnecting existing clients, whenever it's possible.
+   The configuration can be changed dynamically when the server is running (hot reloading) by writing to the configuration file. Changes are detected and applied without disconnecting existing clients, whenever it's possible. Changes are normally detected through filesystem notifications; a periodic check is also performed as a fallback, for setups (for example some Docker bind mounts) where filesystem notifications aren't propagated.
 
 2. Use environment variables, in the format `MTX_PARAMNAME`, where `PARAMNAME` is the uppercase name of a parameter. For instance, the `rtspAddress` parameter can be overridden in the following way:
 
--- a/internal/confwatcher/confwatcher.go
+++ b/internal/confwatcher/confwatcher.go
@@ -12,7 +12,26 @@ import (
 const (
 	minInterval    = 1 * time.Second
 	additionalWait = 10 * time.Millisecond
+
+	// pollInterval is the interval at which the watched file is periodically
+	// stat()ed, as a backstop for setups (for example Docker bind mounts)
+	// where fsnotify never receives events for changes made on the host.
+	pollInterval = 1 * time.Second
 )
+
+// fileState holds the file attributes used to detect changes through polling.
+type fileState struct {
+	modTime time.Time
+	size    int64
+}
+
+func statFileState(path string) (fileState, bool) {
+	info, err := os.Stat(path)
+	if err != nil {
+		return fileState{}, false
+	}
+	return fileState{modTime: info.ModTime(), size: info.Size()}, true
+}
 
 // ConfWatcher is a configuration file watcher.
 type ConfWatcher struct {
@@ -71,6 +90,10 @@ func (w *ConfWatcher) run() {
 
 	var lastCalled time.Time
 	previousWatchedPath, _ := filepath.EvalSymlinks(w.absolutePath)
+	lastState, _ := statFileState(w.absolutePath)
+
+	pollTicker := time.NewTicker(pollInterval)
+	defer pollTicker.Stop()
 
 outer:
 	for {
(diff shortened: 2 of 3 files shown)


aler9:
This adds a permanent ticker and a permanent load to any server instance, regardless of the fact that hot reloading /fsnotify is working or not. It's not the proper way to solve issue #2661, that can be simply solved by configuring the server to point to a file in a mounted folder through `./mediamtx /other-folder/mediamtx.yml`.

PR #6182, opened 2026-09-03
always-available: support disabling recording when the source is offline

### LINKED ISSUE #5587: Add option to prevent 24/7 recording when alwaysAvailable is enabled

xodbs1021:
Supersedes #5767 by @AbdElrahmaN31. Their commit is kept as the first commit of this branch, rebased onto main.
Closes #5587

## Problem

With `alwaysAvailable: yes` and `record: yes`, the offline segment is recorded whenever the source is offline, so recordings fill up with the placeholder picture. #5767 adds `alwaysAvailableRecorded` to skip it. It has been open since May without further activity from its author; I left a review on it with the two gaps below and a suggested fix, and it no longer merges cleanly onto main (`internal/core/path.go`, `internal/conf/conf_test.go`).

## Change

New path parameter `alwaysAvailableRecorded` (default `true`, existing behaviour unchanged). When `false`, recording runs only while a source is connected and stops while the offline segment is playing.

## Reproduction

Config: `alwaysAvailable: yes`, `alwaysAvailableRecorded: false`, `record: yes`, one H264 track. The three regression tests from this branch, run against the original commit alone, show two gaps:

```
--- PASS: TestPathRecordAlwaysAvailableRecorded (4.14s)
--- FAIL: TestPathRecordAlwaysAvailableRecordedStaticSource (10.01s)
        recording did not start after the static source connected
--- FAIL: TestPathRecordAlwaysAvailableRecordedReloadWhileOffline (2.51s)
        recording started during a reload while the source was offline
```

1. Recording was started in `doAddPublisher()` only, so a path with `source:` set (static source) never started recording when its source connected.
2. `doReloadConf()` used `pa.source != nil` as the online check. That is true for a static source even while it is disconnected, so a reload while offline started recording the offline segment.

## Fix

Recording start and stop now live in `setOnline()` / `setOffline()`, which publishers and static sources both go through, and the reload check uses `isOnline()`. Same tests on this branch:

```
--- PASS: TestPathRecordAlwaysAvailableRecorded (4.15s)
--- PASS: TestPathRecordAlwaysAvailableRecordedStaticSource (9.31s)
--- PASS: TestPathRecordAlwaysAvailableRecordedReloadWhileOffline (2.51s)
```

## Impact

- Default `true`: no change for existing configurations.
- `mediamtx.yml` documents the parameter; `api/openapi.yaml` is regenerated with `apidocsgen`.
- Regression tests cover the publisher connect/disconnect cycle, the static source and the reload while offline.


aler9:
merged, thanks!

PR #5779, opened 2026-05-17
feat: support dynamic RTSP proxy paths with regex capture groups (#5766)

OpenCodDev:
- Fixes `bluenviron/mediamtx#5766`
- Enables dynamic RTSP proxy routes with regex captures in `paths`
- Adds support for `$1`, `$2`, ... substitutions in `source` for paths like:
  - `~^mycam/([^/]+)/([^/]+)/(.+)$`
  - `source: rtsp://$1:$2/$3`
- Updates the existing documentation in `docs/2-features/12-proxy.md`
- Includes required code changes in:
  - `internal/conf/path.go`
  - `internal/staticsources/handler.go`
  - `internal/core/path_test.go`

Issue #395, opened 2021-05-21
ERR: username and password must be both provided


eljefe80:
## Which version are you using?

v0.16.1 from dockerhub

## Which operating system are you using?

OS

- [ x] Linux
- [ ] Windows
- [ ] macOS

Architecture

- [ x] amd64
- [ ] arm64
- [ ] arm7
- [ ] arm6

## Describe the issue

I have a couple of no-name ip cameras that require basic auth, without a password. IE, the paths: would look something like:

>   front_room_large:
>     source: rtsp://admin:@10.200.12.53:8001

regrettably, rtsp-simple-server refuses to start saying only:

> rtspproxy    | ERR: username and password must be both provided

The issue appears to be [here](https://github.com/aler9/rtsp-simple-server/blob/cfbea7eedcaca6bdc15bf50ad4372d46ec41607d/internal/conf/path.go#L145)

vlc and ffplay (among other rtsp ingesters) appear to accept this URL format.

## Describe how to replicate the issue

1. configure a paths: entry url with a username, but no password
2. start the server

This failure occurs before network is initiated. 
 no

dorinclisu:
Would the cameras refuse authentication with a dummy password? If not, this would be an easy workaround.

eljefe80:
Yes, using anything but an empty password results in a 401 unauthorized from the camera and I cannot change the password through the interface.

aler9:
Thanks for reporting the issue, sources with empty passwords are now supported in main and will be added to the next release.

aler9:
added to v0.16.2

PR #6193, opened 2026-09-05
conf: allow sources and forward destinations with an empty password

### LINKED ISSUE #5920: v1.19.2 ERR: username and password must be both provided

1.19.2

Linux arm64 standard

### Describe the issue

I believe the same issue as #395 has come back.

```bash
$ ./mediamtx mediamtx.yml
ERR: username and password must be both provided
```

some of my sources are like this:

```
frontdoor_hi:
    source: rtsp://admin:@10.0.0.3:554/h264Preview_01_main
    sourceOnDemand: false
```

Unfortunately setting the log level verbosity in `mediamtx.yml` does not help giving any other clues other than the one line `ERR: username and password must be both provided`

1. start the MediaMTX
2. publish with ...
3. read with ...

### MediaMTX configuration

```yaml
frontdoor_hi:
    source: rtsp://admin:@10.0.0.3:554/h264Preview_01_main
    sourceOnDemand: false
```

### MediaMTX logs

ERR: username and password must be both provided

### Packet dump

_No response_

jant009:
Fixes #5920

## Problem

Since v1.19.1, a source URL with a username and an explicitly empty password is rejected at startup:

```yaml
paths:
  cam:
    source: rtsp://admin:@192.168.1.57:554/videoMain
```

```
ERR: username and password must be both provided
```

This configuration worked from v0.16.2 up to v1.19.0 included. Some legacy cameras (old Foscam models, no-name devices reported in #395) require a username but have an empty password that cannot be changed, so there is no workaround on the camera side.

## Root cause

- In 2021, #395 reported this exact problem and 2bffbe4a deliberately removed the "both provided" check for RTSP sources, adding a `TestSourceRTSPNoPassword` test. The check was kept for RTMP and HLS sources only.
- In v1.19.1, #5779 (regexp groups in source URLs) factored the URL validation of every scheme into a single `validateURL()` function. The RTMP/HLS credential check came along and got applied to RTSP sources too, silently reverting #395. The change was not mentioned in the PR or in the release notes.
- The historical test still exists (`TestNoPassword` in `internal/staticsources/rtsp`) and still passes, but since the 2023 package split it exercises the source layer directly and no longer goes through `internal/conf`, so it could not catch the regression. gortsplib is not involved: v5.5.4 → v5.6.0 has no change in authentication, and its `auth.Sender` handles an empty password correctly.

## Change

`validateURL()` now uses the `ok` flag returned by `url.Userinfo.Password()` to distinguish two cases that were previously treated the same:

| URL | before | after |
| --- | --- | --- |
| `rtsp://user:pass@host` | accepted | accepted |
| `rtsp://user:@host` (explicitly empty password) | rejected | **accepted** |
| `rtsp://user@host` (password missing) | rejected | rejected, clearer message |
| `rtsp://:pass@host` (username missing) | rejected | rejected, clearer message |
| `rtsp://@host` | accepted | accepted |

New error messages:

```
username was provided but password is missing; if the password is intentionally empty, use 'user:@host'
password was provided but username is missing
```

An empty password is valid per RFC 3986 (`userinfo = *( unreserved / pct-encoded / sub-delims / ":" )`), RFC 7617 (Basic) and RFC 7616 (Digest). VLC, FFmpeg and gortsplib all accept it. The original check was only meant to catch a forgotten password, which is still covered by rejecting `user@host`.

Since `validateURL()` is shared, the change applies to every scheme it validates (RTSP, RTMP, HLS, SRT, WHEP, UDP) and to `forward` destinations. gortmplib and the Go HTTP client both handle an empty password.

No configuration option is added: the default behavior stays strict for the typo cases, and the accepted syntax is unambiguous. Client authentication towards MediaMTX (`authInternalUsers`, `internal/auth`) is not touched.

## Tests

- `TestConfFromFile`: new accepted cases for `rtsp://user:@`, `rtsps://user:@`, `rtmp://user:@`, `http://user:@`, `rtsp://@` and `rtsp://user:pass@`.
- `TestConfErrors`: existing `rtsp://user@` case updated to the new message; new rejected cases for `rtsp://:pass@`, `rtmp://user@` and a `forward` destination `rtsp://user@`; new accepted `forward` destination `rtsp://user:@`.
- The existing `TestNoPassword` in `internal/staticsources/rtsp` is unchanged and keeps covering the gortsplib side.
- End-to-end check with a local gortsplib server requiring Digest MD5 with `admin` / empty password: with this patch `rtsp://admin:@127.0.0.1:8554/videoMain` authenticates (DESCRIBE, SETUP, PLAY return 200) and the path becomes ready with one H264 track; a wrong password still yields 401; the unpatched binary still fails at startup with the old error.

## Documentation

`docs/3-publish/08-rtsp-cameras-and-servers.md` now explains that legacy cameras with an empty password must keep the colon (`rtsp://user:@host:port/path`), and that `rtsp://user@host` is treated as a missing password.

### DIFF
--- a/docs/3-publish/08-rtsp-cameras-and-servers.md
+++ b/docs/3-publish/08-rtsp-cameras-and-servers.md
@@ -23,6 +23,8 @@ paths:
 
 If username or password contain special characters (like ?, :, etc), they need to be [url-encoded](https://www.urlencoder.org/).
 
+Some legacy cameras require a username but an empty password. In this case, keep the colon after the username (`rtsp://user:@host:port/path`): a username without a colon (`rtsp://user@host:port/path`) is treated as a missing password and rejected.
+
 The resulting stream will be available on path `/proxied`.
 
 It is possible to tune the connection by using several additional parameters, that are listed in the [configuration file](../5-references/1-configuration-file.md).
--- a/docs/3-publish/08-rtsp-cameras-and-servers.md
+++ b/docs/3-publish/08-rtsp-cameras-and-servers.md
@@ -23,7 +23,7 @@ paths:
 
 If username or password contain special characters (like ?, :, etc), they need to be [url-encoded](https://www.urlencoder.org/).
 
-Some legacy cameras require a username but an empty password. In this case, keep the colon after the username (`rtsp://user:@host:port/path`): a username without a colon (`rtsp://user@host:port/path`) is treated as a missing password and rejected.
+Some cameras require a username but an empty password. In this case, keep the colon after the username (`rtsp://user:@host:port/path`): a username without a colon (`rtsp://user@host:port/path`) is treated as a missing password and rejected.
 
 The resulting stream will be available on path `/proxied`.
 
--- a/internal/conf/conf_test.go
+++ b/internal/conf/conf_test.go
@@ -149,6 +149,30 @@ func TestConfFromFile(t *testing.T) {
 			name:   "udp rtp with placeholders",
 			source: "udp+rtp://$G1:$G2",
 		},
+		{
+			name:   "rtsp with username and password",
+			source: "rtsp://user:pass@localhost/stream",
+		},
+		{
+			name:   "rtsp with username and empty password",
+			source: "rtsp://user:@localhost/stream",
+		},
+		{
+			name:   "rtsps with username and empty password",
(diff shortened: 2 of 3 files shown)


aler9:
merged, thanks!

Issue #5132, opened 2025-10-27
[config] reloading configuration with not exist parent directory of log path will cause panic: runtime error: invalid memory address or nil pointer dereference


bronze1man:
### Which version are you using?

MediaMTX v1.15.3

### Which operating system are you using?

Linux amd64 standard

### Describe the issue

reloading configuration with not exist parent directory of log path will cause panic: runtime error: invalid memory address or nil pointer dereference .
I want it just tell me the parent directory of log path not exist. Or simple just mkdir that dir and make it work. Do not crash the process.

### Describe how to replicate the issue

1. config with logDestinations: [stdout]
2. start the MediaMTX
3. change the mediamtx.yml to logDestinations: [file] and logFile: /opt/mediamtx/log/mediamtx.log and /opt/mediamtx/log not exist.

### MediaMTX configuration

```
# Verbosity of the program; available values are "error", "warn", "info", "debug".
logLevel: info
# Destinations of log messages; available values are "stdout", "file" and "syslog".
logDestinations: [file]
# If "file" is in logDestinations, this is the file which will receive the logs.
logFile: /opt/mediamtx/log/mediamtx.log
```

### MediaMTX logs
```
2025/10/27 09:36:47 INF reloading configuration (file changed)
2025/10/27 09:36:47 INF [API] listener is closing
2025/10/27 09:36:47 INF [WebRTC] listener is closing
2025/10/27 09:36:47 INF [RTMP] listener is closing
2025/10/27 09:36:47 INF [RTSP] listener is closing
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x0 pc=0xf35f60]

goroutine 13 [running]:
github.com/bluenviron/mediamtx/internal/logger.(*Logger).Log(0xc0002cbd40?, 0xc00034ac08?, {0x1646d35?, 0x0?}, {0xc00035a6a0?, 0x18b96c0?, 0xc0003afc80?})
        /s/internal/logger/logger.go:145 +0x40
github.com/bluenviron/mediamtx/internal/core.(*Core).Log(...)
        /s/internal/core/core.go:201
github.com/bluenviron/mediamtx/internal/core.(*Core).run(0xc0002cbd40)
        /s/internal/core/core.go:234 +0x3e5
created by github.com/bluenviron/mediamtx/internal/core.New in goroutine 1
        /s/internal/core/core.go:183 +0x729
[...]

PR #6206, opened 2026-09-06
fix data race between configuration reload and hook logging

xodbs1021:
## Problem

`Core.Log()` reads `Core.logger` while a configuration reload replaces it, without any synchronization.

Since #2278 ([`5a40a06e`](https://github.com/bluenviron/mediamtx/commit/5a40a06e9e5408370c8b7b6c43e38ff16be4facb)), a reload that changes a logging setting closes every subsystem that logs before the logger is replaced, so normally nobody is left to read the field. External command goroutines are the exception: [`externalCmdPool.Close()` is called only when `newConf` is nil](https://github.com/bluenviron/mediamtx/blob/2c6727904fbf233615de74a6c54a9b94dbf6025d/internal/core/core.go#L1327-L1330), that is, at shutdown. Nothing waits for those goroutines during a reload either: `Cmd.Close()` only closes a channel and returns. So when a hook exits, its `OnExit` callback logs through `Core.Log()` and reads `p.logger` while `closeResources()` and `createResources()` are writing it.

The nil half of this was already hit and fixed in #5249 ([`a0a75dfa`](https://github.com/bluenviron/mediamtx/commit/a0a75dfab6ad9094114cdc08a7ea9568451a25e8)) for issue #5132, but only on the `createResources()` failure path; `Core.Log()` itself stayed unguarded.

## Change

- `Core.logger` becomes an `atomic.Pointer[logger.Logger]`, like the `conf` field a few lines above. Every access goes through `Load()` or `Store()`.
- `Core.Log()` loads the pointer once and returns without logging when it is nil.
- Two tests are added, one for each half of the problem: the nil pointer and the race.

## Reproduction

The minimal condition is a reload that changes `logLevel`, plus a `Core.Log()` call while the logger is being replaced. `TestCoreHotReloadingWhileLogging` sets that up with every server disabled, so that the logger replacement is the only thing that runs concurrently with the logging. On a tree with `internal/core/core.go` reverted to `main`, and the test's two `p.logger.Load()` calls changed to `p.logger` since a plain pointer has no `Load()`:

```
$ go test -race -count=1 -run TestCoreHotReloadingWhileLogging ./internal/core
==================
WARNING: DATA RACE
Write at 0x00c0003d6238 by goroutine 12:
  github.com/bluenviron/mediamtx/internal/core.(*Core).closeResources()
      internal/core/core.go:1336
Previous read at 0x00c0003d6238 by goroutine 13:
  github.com/bluenviron/mediamtx/internal/core.(*Core).Log()
      internal/core/core.go:376
```

Run in 20 separate processes, that command failed 20 times out of 20 on macOS/arm64; every run also carried a second `DATA RACE` block, for the write in `createResources()`, and a nil dereference panic. Inside `golang:1.26-alpine3.24` on Linux/arm64 it failed 20 out of 20 the same way. `TestCoreLogWithoutLogger` covers the nil pointer and needs no race detector: on the same reverted tree it failed in all 3 runs, on the nil dereference inside `logger.Logger.Log()`. On this branch both tests pass:

```
$ go test -race -count=1 -run 'TestCoreLogWithoutLogger|TestCoreHotReloadingWhileLogging' ./internal/core
ok  	github.com/bluenviron/mediamtx/internal/core	3.920s
```

## Fix

The field becomes an atomic pointer rather than a mutex, for two reasons. The `conf` field is already one and `sync/atomic` is already imported, so no new concept enters the struct. And every subsystem's log entries end up in `Core.Log()`: entries that the level filter throws away would pay for the lock too, since that filter runs after the logger has been read.

A mutex that only guards the swap would end up doing the same thing: the reader takes the lock, sees no logger, and drops the entry. Making it lossless means holding it across `oldLogger.Close()` too, which serializes the reload against every log call in the process.

## Impact

Logging behavior is unchanged while a logger is installed. Entries produced while no logger is installed are discarded, where before they raced with the replacement or dereferenced a nil pointer.

The atomic pointer also closes unsafe publication, not only the nil dereference: without the release/acquire edge that `Store()` and `Load()` give, a reader can see the new `logger.Logger` pointer before the writes that fill it. A plain pointer with an `if p.logger != nil` check around the call still reported a race in 20 runs out of 20 on both platforms, between the read of `Level` inside `Logger.Log()` and those writes in `createResources()`. A few of those runs also panicked outright, because that form loads the field twice and the value can turn nil between the check and the use.

One pre-existing case is left as it is: `reloadConf()` closes the old logger, so an entry that loaded the pointer just before can still be written to a logger that is already closed. This PR neither creates that path nor widens it.

### DIFF
--- a/internal/core/core.go
+++ b/internal/core/core.go
@@ -207,7 +207,7 @@ type Core struct {
 	confPath        string
 	conf            atomic.Pointer[conf.Conf]
 	supportsIPv6    bool
-	logger          *logger.Logger
+	logger          atomic.Pointer[logger.Logger]
 	externalCmdPool *externalcmd.Pool
 	authManager     *auth.Manager
 	metrics         *metrics.Metrics
@@ -346,7 +346,7 @@ func New(args []string) (*Core, bool) {
 
 	err = p.createResources(true)
 	if err != nil {
-		if p.logger != nil {
+		if p.logger.Load() != nil {
 			p.Log(logger.Error, "%s", err)
 		} else {
 			fmt.Printf("ERR: %s\n", err)
@@ -373,7 +373,12 @@ func (p *Core) Wait() {
 
 // Log implements logger.Writer.
 func (p *Core) Log(level logger.Level, format string, args ...any) {
-	p.logger.Log(level, format, args...)
+	// a configuration reload can remove the logger before installing the
+	// new one, and a shutdown removes it for good; entries produced while
+	// no logger is installed are discarded.
+	if l := p.logger.Load(); l != nil {
+		l.Log(level, format, args...)
+	}
 }
 
 func (p *Core) run() {
@@ -398,7 +403,7 @@ outer:
 		case <-confChanged:
 			p.Log(logger.Info, "reloading configuration (file changed)")
 
-			newConf, _, err := conf.Load(p.confPath, nil, p.logger)
+			newConf, _, err := conf.Load(p.confPath, nil, p.logger.Load())
 			if err != nil {
 				p.Log(logger.Error, "%s", err)
 				break outer
@@ -500,7 +505,7 @@ func (p *Core) createResources(initial bool) error {
 	currentConf := p.conf.Load()
 	var err error
 
-	if p.logger == nil {
+	if p.logger.Load() == nil {
 		i := &logger.Logger{
 			Level:        logger.Level(currentConf.LogLevel),
 			Destinations: currentConf.LogDestinations.ToDestinations(),
@@ -512,7 +517,7 @@ func (p *Core) createResources(initial bool) error {
 		if err != nil {
 			return err
 		}
-		p.logger = i
+		p.logger.Store(i)
 	}
 
 	if initial {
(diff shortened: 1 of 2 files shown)


aler9:
The reported bug is correct but the proposed fix is too performance-heavy and too different from the rest of the code, basically it acquires a mutex for each logged line. All of this for the borderline case of destroying and recreating the logger. It would be more appropriate, in that case, to close everything (including `externalCmdPool`) in order not to have anyone using the logger, and then recreate the logger.

xodbs1021:
@aler9 Thanks, I redid the fix the way you suggested.

`externalCmdPool` is now closed on a reload that recreates the logger, before the logger is replaced, and `createResources()` creates it again; `Core.Log()` is untouched and the diff in `core.go` is six added lines and one removed; the other components that depend on the logger were already closed and recreated on such a reload.

The consequence is that such a reload now waits for the running hooks to exit on SIGINT: measured as the interval in which the RTSP listener is unbound, 0.25 ms on `main` against 3.01 s here with a hook that exits 3 s after the signal, and with no hook the difference is within the noise of the measurement. If you would rather bound that wait, I can change it. A hook that does not exit on SIGINT keeps the reload waiting, as it already keeps shutdown waiting.

The new test drives real hooks; on `main` it fails without the `core.go` change and passes with it.

Issue #6247, opened 2026-09-20
WebRTC: optionally associate KLV messages with video RTP timestamps


adamdayan:
MediaMTX currently sends raw KLV payloads over a WebRTC data channel without their presentation timestamps. Consequently, once video and KLV are carried separately over RTP and SCTP, clients have no reliable way to associate a KLV message with its corresponding decoded video frame. Falling back to arrival-time matching is sensitive to buffering, scheduling, and network jitter.

MediaMTX already has both the KLV presentation timestamp and the outgoing video RTP timestamp, so it can preserve this association before the streams diverge.

I propose an opt-in global setting:

  webrtcKLVDataChannelFormat: raw # raw | timed

  - raw remains the default and preserves the existing KLV data channel.
  - timed exposes a KLV-TIMED channel containing a small versioned envelope:

```
  version:             uint8
  flags:               uint8
  header_length:       uint16
  video_rtp_timestamp: uint32
  klv_payload:         remaining bytes
```

adamdayan:
I've had a first go at implementing this [here](https://github.com/bluenviron/mediamtx/pull/6248) - very happy to take feedback or suggestions on how to implement this better!

PR #6248, opened 2026-09-20
Add optional timing metadata to WebRTC KLV

adamdayan:
MediaMTX currently sends KLV over WebRTC without presentation timing, leaving clients to associate metadata with video frames by arrival time.

This initial proposal adds an opt-in timed KLV data-channel format containing the corresponding outgoing video RTP timestamp. The existing raw format remains the default and unchanged.

This is only a first go at solving #[6247](https://github.com/bluenviron/mediamtx/issues/6247), very happy to take feedback or suggestions!


adamdayan:
@aler9 would you mind enabling a CI run? thank you!

aler9:
GitHub doesn't allow me to unlock CI due to too many conflicts.

Anyway, the fact is that in RTP (used by both WebRTC and RTSP) timestamps are supposed to have a random initial offset, and peers are supposed to detect these initial offsets and and remove them from timestamp of subsequent packet in order to get some sort of absolute timestamp. Is it precise? absolutely not, because the resulting timestamp is influenced by network fluctuations and system clock fluctuations.

The suggested way to ship a precise and absolute timestamp in RTP consists in using an extension that allows to write the absolute timestamp in some bytes at the beginning of each packet:
https://www.ietf.org/archive/id/draft-alvestrand-avtcore-abs-capture-time-00.html

And i would use that, for both video and KLV.

adamdayan:
thanks for the quick response @aler9 !  I didn't know about that Absolute Capture Time RFC - absolute capture time does seems preferable when the source provides an authoritative absolute time. However, there are a few places I'm still confused:
 - the KLV is going over SCTP which doesn't have a timestamp field. Are you suggesting that I keep the KLV timestamped envelope change but just write the Absolute Capture Time into it? 
 - I understand how your proposal would work for RTP with the ACT, but I don't think it would work for muxed MPEG-TS video + KLV as MPEG-TS has no generic absolute time field. In the MPEG-TS case, it seems like ensuring the KLV uses comparable PTS to the video is the best we can do? 

I focussed too much on the MPEG-TS case where there is no absolute time, and KLV and video share the same PES PTS timeline. Would a reasonable compromise that handles both cases be using Absolute Capture Time where it's available, and otherwise use PTS correlation and signpost this with a flag in the KLV timing envelope?

adamdayan:
@aler9 apologies for the disturbance, any view on [these questions](https://github.com/bluenviron/mediamtx/pull/6248#issuecomment-5768559472)? thank you!

Issue #6250, opened 2026-09-21
Join UDP Multicast Groups On All Network Interfaces


benjamin-knight-sai:
### Describe the feature

I am looking for a way to specify joining UDP multicast groups on _all_ network interfaces instead of the default one or a specific one. 

It seems this used to be the default functionality, but got updated in this commit https://github.com/bluenviron/mediamtx/pull/4820/commits which I believe is related to this issue https://github.com/bluenviron/mediamtx/issues/4565.

In the issue, the author brings up the possibility of implementing this:

> MediaMTX's current behaviour of trying to get video from all interfaces could then be set as follows:
`source: udp://238.0.0.1:1234?interface=ALL`
Although someone might be upset that it's a change in behavior.  

But it appears that never made it in. Is there something I may be overlooking with this request? I understand the risk of there possibly being two producers on different interfaces which would break the video consumption but that's why this should be an explicit flag to toggle on the behavior.

ben8622:
pull request made for this functionality https://github.com/bluenviron/mediamtx/pull/6253

PR #6253, opened 2026-09-21
udp: support `interface=all` for multicast sources 

ben8622:
(related to this issue https://github.com/bluenviron/mediamtx/issues/6250)

## Summary

This PR adds support to listen for a UDP multicast group on all multicast-capable network interfaces. To do so, add the `interface=all` parameter to the end of a UDP multicast URL when configuring the source. For example:
```
udp+mpegts://238.0.0.1:1234?interface=all
```

## Use Case

1.) Finding the network interface via default logic, ([code link](https://github.com/bluenviron/mediamtx/blob/main/internal/protocols/udp/listener.go#L22-L58)), for a UDP multicast source can  be incorrect if someone forces specific routing rules for the multicast group's address.  

2.) If the network interface of the UDP multicast source can _change_ during the lifetime of MediaMTX I want it to be able to react without having to manually re-configure the path's configuration.  

## Possible Risks

- Conflict with an actual network interface called `all`, I believe this is very unlikely to happen as you would need some custom configuration and setup to do this.
- If there are two producers sending multicast traffic to the same group on different network interfaces this listener would consume both which would produce a corrupted video feed. This is definitely possible, but that is why you would have to explicitly toggle this feature on. 

## Testing 
Added a test case validating when `all` interface is supplied to the listener's `Initialize()` function that we are using `NewMultiConn` instead of `NewSingleConn` and leave the actual interface logistics to the `gortsplib` library that implements those classes. True testing of this feature is probably more involved then needed here since it would require having two multicast capable interfaces, spinning up video feed on the non-default one, and testing out we receive traffic when supplying `interface=all`.


ben8622:
I was able to test this for my specific use case and it works as expected.

- Computer A, hosts MediaMTX server, has several different multicast capable network interfaces but lets focus on just two WiFi and a LAN.
- Computer B, published video feed over UDP multicast, connected to Computer A over LAN
- Computer A has two MediaMTX paths configured:
```
# source one
"source": "udp+mpegts://239.1.1.1:5000"
...
# source two
"source": "udp+mpegts://239.1.1.1:5000?interface=all"
```
- We can see source one defaults to the WiFi network interface and fails to consume any traffic
- Source two joined the group on all interfaces then received traffic which allowed me to view the video as normal

